PR #16, opened 2017-04-25
Refactored ReadableFloatVector

lukasbeck:
Refactored `ReadableFloatVector` by making method `getEuclideanDistance` static and providing one general implementation.


lukasbeck:
Please do not merge yet, I would like to make both `toArray` and `toList` static as well.

lukasbeck:
Should be fine now. Let me know if something is amiss.

PR #18, opened 2017-04-25
Added SpatialDistance feature

lukasbeck:
Added `SpatialDistance` feature that calculates similarity based on an approximation of the great-circle distance between two points given in `(latitude, longitude)`. The feature serves as a `MetadataExtractor` and tries to extract given location information from a document. That is, it uses EXIF, if available. Otherwise it tries to retrieve the coordinates from an additional JSON file with the same name as the document (e.g. `image_1337.json` for `image_1337.jpg`), given as keys in the JSON file, e.g.:

    {
      "latitude": 42.03,
      "longitude": 4.47
    }

Main changes:
* Added `SpatialDistance`, `Location` and `JsonMetadataExtractor`
* Extended `MetadataExtractor` to support initialization & finishing, adapted `AbstractExtractionFileHandler` accordingly.
* Segments may now contain a location. This is needed in order to enable location-based queries. It is an unfortunate consequence of queries being represented as `SegmentContainer`s.


lucaro:
Since the behavior of `SpatialDistance` is not immediately obvious from its name, concerning where it gets its information from (especially with the json-fallback), having a bit of documentation would be nice. Otherwise I don't see anything preventing a merge.

lukasbeck:
I extracted an abstract super class `MetadataFeatureModule` out of `SpatialDistance` to make it easier for future features using objects instead of segments (e.g. temporal). I also added several comments for both classes, in particular describing the difference regarding segment vs. object extraction & retrieval as well as how `SpatialDistance` extracts the location information.

PR #40, opened 2017-11-22
Adding Logic to retrieve for one single feature by Name

silvanheller:
Could be easily extendable to query for one retriever instance


lucaro:
Why is only id-based lookup supported?

silvanheller:
Now supports lookup by id, QueryContainer, Retriever-Name and Retriever-Instance

PR #53, opened 2018-04-17
Generic Extraction

### LINKED ISSUE #50: Refactor AbstractExtractionFileHandler to enable multiple media types per extraction-session
Once #48  is merged, this is the next item on the wishlist
Some thoughts:
- [ ] Move Executor-Service instantiation out of the FileHandler so it can be shared between multiple FileHandlers.
- [ ] ExtractionItemContainers are requested with specified media-type support OR are checked for compatibility from a pool of filehandlers 
- [ ] I'm not sure if the FileHandlers can share an ExtractionPipeline.

silvanheller:
# Generic Extraction  - Resolves #50 

# Feature Overview
* The new GenericExtractionItemHandler is very similar to the AbstractFileHandler, but has a map of mediatypes and their corresponding decoders / segmenters.
* There is a new Interface, the ExtractionItemProcessor. This decouples the ExtractionDispatcher from the AbstractFileHandler.
* Removed the ExtractionFileHandler interface and moved its capabilities to the AbstractFileHandler.
* Performing an endSession() call does not close the extractionpipeline instantly anymore, but instead signals to the provider that it should close itself after it has handed out all paths. This massively improves testability.

# Performed Tests
* Audio Works (Visual Features do not generate anything)
* Images Work (Audio Features do not generate anything)
* Video Works
* Extractionsession with audio, video and images.

# Other Improvements
* Javadoc cleanup
* Logging Improvements
* toString() Methods
* Fixed Multimedia
* Upgrades Gradle to 4.7 in Preparation of Java 9+ compatibility (We are currently limited to Java 8 because grpc generated code uses annotations which are not publicly visible)
* Monitoring Fixes for Mac: Moved from localhost to docker.host.internal since this is the preferred way of accessing the host NW on mac.


src/org/vitrivr/cineast/core/data/MediaType.java
```diff
@@ -16,6 +16,7 @@
     IMAGE(1, "i", "image"),
     AUDIO(2, "a", "audio"),
     MODEL3D(3, "m", "3dmodel"),
+    ALL(98, "l", "all"),    //Used for legacy compatibility in job-scripts
```
lucaro: I'm not sure about this one, it might cause issues when an objectdescriptor of this type gets persisted for some reason. Wouldn't it be sufficient to make the parameter in the job file optional? Having it not set would be equivalent to setting it to 'all'.

src/org/vitrivr/cineast/core/data/entities/MultimediaMetadataDescriptor.java
```diff
@@ -103,7 +104,14 @@ public MultimediaMetadataDescriptor(
     if (value != null & isSupportedValue(value)) {
       this.value = new StringTypeProvider(value.toString());
     } else {
-      this.value = new NothingProvider();
+      if (value != null && value.getClass() == StringTypeProvider.class
+          || value.getClass() == StringProviderImpl.class) {
+        this.value = new StringTypeProvider(((StringProviderImpl) value).getString());
```
lucaro: Checking `value instanceof StringProvider` should be equivalent while also covering other potential future sub-classes.

src/org/vitrivr/cineast/core/data/MediaType.java
silvanheller: Implemented in 5c45630b685797e8deed1d3364cdaad80ac61517

PR #65, opened 2018-07-18
Implemented Add Media API.

iammvaibhav:
Add Media API is completely implemented. This implementation is different from the one I did previously. Main highlights are:

1) No need to start the cineast server with --server option along with a config file. When a request arrives, provider will be initialized/opened if closed, extraction will take place and then session will be ended so as to write data to ADAMPro.

2) Instead of starting cineast server with --server config, there must be 4 config files (for image, video, audio, 3d model) in the same directory as cineast.jar. This is done mainly because of three reasons.
i. If there is only one config files, all the exporters need to be specified (ShotThumbNails, AudioWaveformExporter, Model3DThumbnailExporter etc) for all different media types. The problem with this is if say an image is getting processed, and if AudioWaveformExporter is specified, it will process and export the waveform even though it doesn't make any sense for an image. 
ii. In case of single config files, only a single set of extractors can be specified for all types even when say LightfieldZernike doesn't make any sense while extracting an image. This doesn't necessary creates any problem except taking some extra time in checking. When using multiple configs, different set of extractors can be specified for different media types. 
iii. In case two or more media type uses a same exporter (ex both IMAGE and VIDEO can use ShotThumbNails) then they will be exported to the same directory (which is specified in config). This creates problems when retrieving thumbnails in current client scenario which expects thumbnails of different media types to be in different directories. When using multiple config files, this problem will also be solved.
Also if required, request can specify the extractors and exporters to be used while extraction. This isn't implemented yet as I'm not very sure about it's utility.
The current implementation requires image_extraction_config.json, video_extraction_config.json, audio_extarction_config.json, and 3dmodel_extraction_config.json in the same directory as cineast.jar each having input (for Unique id generator), extractors and exporters block.

3) Users can upload mutiple files at once of the SAME type (viz IMAGE, VIDEO, AUDIO, MODEL3D) along with the metadata if any. Specifically
i. Request will be of type multipart/form-data.
ii. A header media_type needs to be specified which tells the media type of all the files.
iii. files to be extracted should be uploaded with filename as their key.
iv. A key extract_config should be specified along with the file which will be deserialized to 'ExtractionItemContainer'.
v. In the object section of ExtractionItemContainer, name and path fields should be set to file name and uri can be any valid dummy uri (because of URI validations eg file:///dummy)

4) Valid objectlocation needs to be specified in api section of cineast.json. The uploaded files will be segregated according to their media types so as to comply with the current client specifications.

### DIFF
--- a/src/org/vitrivr/cineast/api/APIEndpoint.java
+++ b/src/org/vitrivr/cineast/api/APIEndpoint.java
@@ -164,6 +164,7 @@ public class APIEndpoint {
             service.post("/extract/end", new EndExtractionHandler());
             service.post("/extract/start", new StartExtractionHandler());
         });
+        service.post(makePath("extractFiles"), new FileExtractionHandler());
         if (Config.sharedConfig().getApi().getServeContent()) {
           service.path(makePath("get"), () -> {
             service.get("/thumbnails/:id", new ResolvedContentRoute(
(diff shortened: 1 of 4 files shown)


silvanheller:
As mentioned in our conversation offline, we'll merge this into a gsoc-2018 branch for submission

PR #66, opened 2018-08-04
Restored consistency in certain query responses

sauterl:
TL;DR Websocket and RESTful API now respond with the same JSON object's and lists are properly enclosed by an object.

The following is a copy of  sauterl/cineast@940f74c2ee13d78b8427c3de214e611189e23249 describing the consequences of the change:

---

Restored consistency in certain query responses

The WebSocket and RESTful API now returns the same JSON objects for

 * /find/metadata/by/id
 * /find/objects/by/id
 * /find/segments/by/id

Each of the corresponding RESTful action handler's now do have the same result
type as the corresponding WebSocket handlers have.

The main advantage is that both APIs do now have similar response
objects.
Further, there is a minor advantage in not sending top level arrays: A
few JSON processors cannot handle such documents and would therefore not
be able to be used in a cineast context.


sauterl:
I suggest to review the responses of the entire RESTful API, as I recently realised, that there are other actions without top-level objects (see #69  ).

ppanopticon:
The change is fine from my perspective. However, some minor adjustments to Vitrivr NG will be required.

I agree with Loris that maybe we should put some effort into unifying the responses of both the RESTful and the WebSocket API.

lucaro:
Very well, in that case I will merge this for now.

Issue #63, opened 2018-07-16
[Wishlist] Filter Metadata


sauterl:
It would be very convenient if one could filter metadata domains at request-time:

To illustrate the aforementioned feature, let's have a look to the following example: 

When requesting metadata of several objects (for RESTful API), the request to `/api/v1/find/metadata/by/id` would may be look like:

```
{
  "filters": [
   "LOCATION, "JSON"
   ],
  "ids":[
    "i_zD0BXwe4a7ljNbRJ",
    "i_1bBhZC4JLdf7RGKL"
  ]
}
```

The response would only contain metadata entries of the domains `LOCATION` and `JSON`

silvanheller:
#68 seems to resolve this

Issue #64, opened 2018-07-16
[Wishlist] Sophisticated Metadata API


sauterl:
As an attendum to #63 

In general more cineast-side filtering / processing of metadata would be highly appreciated:
This includes, but is not limited to:

 * specify domain on request-time (see #63 )
 * specify domain and key in the request (e.g. for a GET request: `api/v1/metadata/of/:objectID/:domain/:key` )
 * by domain (e.g. for a GET request: `api/v1/metadata/in/:domain` )
 * by key, regardless domain (e.g. for a GET request: `api/v1/metadata/with/:key`)

silvanheller:
#68 seems to resolve this

PR #68, opened 2018-08-04
Feature extended metadata api

sauterl:
This PR addresses and resolves vitrivr/cineast#63 and vitrivr/cineast#64.

In general, the changes in this PR lead to a more sophisticated metadata api for the RESTful API, as the Websocket API was not touched (yet).

Three new RESTful actions where added and an existing one was slightly altered, as the remainder of this desription describes:

## Altered `find/metadata/by/id`

Optionally there are filter's allowed in the JSON body.
Currently two filters are supported: `domain` and `key`.
The former filters on the domain property, the latter on the key property of metadata.

Given the following example JSON body:

```
{
  "filters":[
    {
      "type":"domain",
      "keywords":[
        "location"
      ]
    },
    {
      "type":"key",
      "keywords":[
        "latitude"
      ]
    }
  ],
  "ids": [
    // some ids
  ]
}
```

This will result in a response containing only metadata of domain _location_ (case insensitive) and with the key _latitude_ (case insensitive) for the given ids.

Adding several filters of the same type is not recommended, as they are applied sequentially (first in, first out). All filtering is case insensitive.

Omitting the filters array results in an unfiltered metadata list response.

## Domain based metadata

Enables domain-based metadata lookup, where the client does not have to filter metadata any more.

### GET

The new action `find/metadata/in/:domain/by/id/:id` results in a list of metadata from the domain _:domain_ for the object with id _:id_.

### POST

The new action `find/metadata/in/:domain` will result in a list of metadata from the domain _:domain_.
The request body is expected to be an IdList.

## Key based metadata

Enables key-based metadata lookup, regardless of the domain, again having the advantage of server-side filtering.

### GET

The new action `find/metadata/with/:key/by/id/:id` results in a list of metadata with the key _:key_

Edit: ~**NOTE: As of a94eb4d, this feature is not yet implemented**.~ This feature exists as of a4f9bd2.

### POST

The new action `find/metadata/with/:key` results in a list of metadata with the key _:key_.
The request body is expected to be an IdList.

## Specific metadata item

Sometimes it is desired to get specific metadata of one or multiple objects.
This new action enables this feature.

### GET

The new action `find/metadata/of/:id/in/:domain/with/:key` results of the metadata item of the object with the specified _:id_, _:domain_ and _:key_.

### POST

This is roughly covered by the optional filters of action `find/metadata/by/id`

### DIFF
--- a/src/org/vitrivr/cineast/api/APIEndpoint.java
+++ b/src/org/vitrivr/cineast/api/APIEndpoint.java
@@ -108,6 +108,7 @@ public class APIEndpoint {
         /* Configure the result after processing was completed. */
         service.after((request, response) -> {
             response.header("Access-Control-Allow-Origin", "*");
+            response.header("Access-Control-Allow-Headers", "*");
         });
 
         return service;
--- a/src/org/vitrivr/cineast/api/APIEndpoint.java
+++ b/src/org/vitrivr/cineast/api/APIEndpoint.java
@@ -151,6 +151,7 @@ public class APIEndpoint {
             service.post("/objects/by/id", new FindObjectByActionHandler());
             service.post("/metadata/by/id", new FindMetadataByObjectIdActionHandler());
             service.post("/metadata/in/:domain", new FindMetadataInDomainByObjectIdActionHandler());
+            service.post("/metadata/with/:key", new FindMetadataByKeyByObjectIdActionHandler());
             service.post("/tags/by/id", new FindTagsByActionHandler());
         });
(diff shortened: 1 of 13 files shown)


lucaro:
This now conflicts with dev after merging #67, @sauterl could you please have a look at this?

sauterl:
The merge conflicts were caused due to the fact, that this branch already included the changes of #66 (at least in the metadata action handler).
Should be fixed by f54e301 and was tested using a CURL-Client (insomnia).

PR #76, opened 2019-05-15
Bugfixes for LSC19 which also belong to dev

silvanheller:


lucaro:
a57fc32e597a292f9e68757ecd5401a051471d61 does not really belong into that list...

PR #77, opened 2019-07-09
LSC 19 final merge

silvanheller:
This merges all important changes from lsc19 to dev. Theses include mostly small things + the upgrade to the newest cottontail-version.


silvanheller:
> With the exception of the mess that is 'cineast.json', this PR looks fine. It looks like it contains all relevant changes made in the lsc19 branch.

It should look nicer now, i replaced all tabs with two spaces.

PR #78, opened 2019-07-16
Cineast 2.1

silvanheller:
# Cineast / Vitrivr 2.1 (~ What we used at LSC19)

* Corresponding PR on the vitrivr-ng repo: https://github.com/vitrivr/vitrivr-ng/pull/13

This PR is happening for the following reasons:
* dev/ is probably **more** stable than master/ since it has received a lot of attention and testing since 2017 and most bugfixes have not been ported to master
* The pattern of keeping all features on dev/  and not updating master/ is unhealthy for an open source project.
* The last real change on master happened in October **2017** This does not reflect the activity around vitrivr since then and gives off a bad impression for people visiting the repository
* The current setup guide on vitrivr.org is based on the master branch, which is not ideal for new users, existing users and student projects

# Changelog
* Cineast now uses [CottontailDB](https://github.com/ppanopticon/cottontaildb) instead of ADAMpro by default, which has resulted in massive speedups for all query types and modes 
* Boolean retrieval (LSC 19)
* New _add media API_ which can be used by mobile apps or other systems (GSOC 18)
* Monitoring via Prometheus / Grafana
* Improved metadata retrieval
* New feature: Semantic Sketching (VBS19)
* New feature: Image Fingerprinting
* Improved text retrieval (VBS19)
* Improved streaming of query results
* No need to specify filetype anymore during extraction
* Various bugfixes and other improvements


lucaro:
> Optional: Remove both org.vitrivr.cineast.explorative.* and org.vitrivr.cineast.art.* from the codebase since they are unsupported / without owner

As sad as it is, but since there is no API to this functionality anymore and it is not supported by the current iteration of the UI anyway, it might just as well be removed. Maybe there is an opportunity to add similar functionality again in the future.

> Optional Restructure the repo layout to be consistent with gradle guidelines (would also simplify our build.gradle file

I would leave this for v3, which will probably bring some major restructuring anyway. As far as I know, @ppanopticon already started something along those lines.

ppanopticon:
Go4it

PR #92, opened 2019-10-30
Multi module to dev

lucaro:
Since there appears to be agreement among all current contributors that the new multi-module project structure is the way to go and the multi-module branch has therefore become the de-facto dev branch, I suggest to make it official...


silvanheller:
Very much in favor of this

ppanopticon:
Go for it!

PR #107, opened 2020-05-26
avoid getting first k of batched knn only

mpas97:
A batched knn query is designed to return k*n results. Til now, k was set as query limit, so we always got the k nearest neighbours of the first provided segment only.

### DIFF
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailSelector.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailSelector.java
@@ -87,7 +87,7 @@ public class CottontailSelector implements DBSelector {
             vectors,
             null,
             k,
-            configs.get(0).getDistance().orElse(Distance.manhattan)), k);
+            configs.get(0).getDistance().orElse(Distance.manhattan)), null);
 
     List<QueryResponseMessage> results = this.cottontail.query(CottontailMessageBuilder.queryMessage(query, configs.get(0).getQueryId().toString()));
 


silvanheller:
@ppanopticon can you briefly confirm that this change should yield the desired behavior?

ppanopticon:
Yes! This should actually fix the issue!

However, I'm not sure if instead we should simply do away with the LIMIT-clause all together in CottontailMessageBuilder.query() (lines 191 to 193), since kNN queries are inherently limited by the k which renders this clause redundant.

sauterl:
To quickly jump into the discussion. I would favor the "kNN is inherently limited" aaproach @ppanopticon suggested. To me, it appears more intuitively

mpas97:
I'm not exactly sure if that's what @ppanopticon and @sauterl have meant here.
Anyway, with the last commit our limit in CottontailMessageBuilder.query() became optional now.

lucaro:
I object to the implementation using varargs, since it implies a very different semantic. Why would I want to specify an arbitrary amount of values here. This looks to me like a lazy way of overloading a method. If the parameter is supposed to be optional, overload the method properly, make the parameter an Optional or at least nullable.

mpas97:
It is, that's why I was confused doing this. What do you mean by making it nullable? As you can see in the first commit, we could always just pass null to ommit the limit. Or am I missing any clue here? Because none of the arguments is an Optional although we explicitely allow null values in almost any of them. Annotations like `@Nullable` don't seem to be used eighter. So just changing one parameter to some sort of optional if all others can have null too, doesn't make much sense to me. Eighter we change everything or nothing, else it's much more confusing than this simple and consistent null of the beginning.

lucaro:
In the case of a primitive like a number, to which you cannot assign `null` since they are not objects, the boxing types need to be used to be able to assign a `null` value. This was already the case previously, using `Integer` rather than `int`. Even though this is not an elegant solution, it is a more intuitive one. Using Optionals would be a more proper way to do this (since java does neither support default nor named parameters), but since it is not done so in the rest of this part of the API, I would vote for consistency and the avoidance of additional boilerplate.

mpas97:
Exactly. Having added an overloaded method now (to avoid Optionals), seems the best solution in this context for maintaining consistency.

lucaro:
Technically, overloading the method would mean to have it twice, once with and once without the parameter. But having the parameter nullable is fine in this case.

mpas97:
Or would you even remove that overloaded method again? As said, we could still just provide null to the original method. Because this way, if we want to be consistent, we would even have to write an overloaded method for each of the possible parameter configurations. What if we, e.g., don't want to provide a knn? Either we leave the original where you just set null to those you don't need, or we overload every combination...

mpas97:
I just missed the last comment as I was writing mine. But so I will revert everything to the state of the first commit.

silvanheller:
Thanks for the fix!

PR #109, opened 2020-05-31
Fixed OpenAPI route specifications not being documented properly

SamuelBoerlin:
Just a small change so that the OpenAPI route specifications documentation works again. Since the `@OpenApi` annotation is applied to the implementation methods (`doGet/doPut/etc.`) the documentation data needs to be copied from that annotation to the actual handlers (`get/put/etc.`), so that the Javalin OpenAPI plugin can know about it.


SamuelBoerlin:
I just realized that this doesn't work e.g. for `ParsingPostRestHandler`'s because they have the annotation on another method. I'll make a new PR once/if I find a better solution.

PR #110, opened 2020-07-06
Upgrade gradle to version which supports JDK 14+

frankier:
See https://github.com/gradle/gradle/issues/10248


lucaro:
Hey, there are a lot of changes underway to the build environment in #104 which is planned to be merged soon(tm). Any updates to gradle should therefore target dev. How unhappy is gradle after this version update in terms of deprecated instructions which are still used in the build script?

frankier:
OK. I will base everything off of `dev`. The newer version of gradle seems to be working fine so far. The only things I've had are warnings, probably as a result of changes between OpenJDK 8 and 14. See: https://singularity-hub.org/containers/13554/log -- but that's a separate issue.

PR #112, opened 2020-07-09
Create directory in CachedDataFactory lazily

frankier:
This stops CachedDataFactory always attempting to create a directory in
the current working directory as a result of the construction of a
default instance. This will fail if that happens to be a read only file
system, as e.g. with Singularity.


frankier:
Actually if I recall correctly, it crashes with a NullPointerException when trying to log the fatal exception. I didn't get around to figuring out what was null. For me, it's a lot more useful that it actually died since it means I notice the error faster. I suppose the reason why Cineast always tries to continue is because usually it's running as a server and should try to continue to service other requests, but in the case of extractions, which are typically batch processes I think fatal problems should tear everything down (I also went into this a bit here: https://github.com/vitrivr/cineast/issues/114 ).

silvanheller:
Good point. I guess there's a larger discussion to be had around failing gracefully vs failing fast, but for the time being, this fixes a reproducible issue so I'm merging it. Thanks for your work!

Issue #79, opened 2019-08-12
QoL: Create Table for Extractors when starting a job if they are not present already


silvanheller:

silvanheller:
#114 is a better place to discuss default behavior when the database is either unavailable or not set up the way it should be. Currently, errors are shown to the user and the setup command is documented.

PR #117, opened 2020-07-23
Allow extraction config to be passed to specify database entities to setup

frankier:
This patch allows --extraction to be passed to setup to setup tables/entities from the extraction json instead of the cineast json, which could be convenient e.g. if you're doing a batch extraction of many features and you might expose a subset later on.


frankier:
I suppose this partially addresses this https://github.com/vitrivr/cineast/issues/79

Issue #115, opened 2020-07-12
Extraction is bottlenecked / doesn't make full usage of multiple cores


frankier:
When performing an extraction on a machine with 40 cores and the following settings:

```
  "extractor": {
    "threadPoolSize": 64,
    "taskQueueSize": 80,
    "shotQueueSize": 80,
    "outputLocation": "out"
  },
```

Most of the resources as not used. This is using the JSON writer as advised for HPC extraction in the other issue.

About 70% of the time it's as if single threaded
![Screenshot from 2020-07-11 14-54-19](https://user-images.githubusercontent.com/299380/87224970-20229400-c392-11ea-9641-418271fadfea.png)

About 30% of the time it uses slightly more but not close to the full 40 cores
![Screenshot from 2020-07-11 14-55-40](https://user-images.githubusercontent.com/299380/87224969-1e58d080-c392-11ea-8f6e-44847d7e9fc3.png)

Since this is essentially an embarrassingly parallel problem (each video is independent), we should be able to do improve upon this which would help for performing extraction on large collections.

An easy way of fixing this would be to have "external" parallelism where many extraction processes are started with Snakemake or similar (I have been using this together with slurm like so https://github.com/frankier/singslurm ) and each extraction process would only deal with a single file before finishing, which would also allow extraction across multiple nodes. The resulting JSONs could then be merged later on. This is already almost possible I think by templating the file name into the the extraction job config but in this case I think the paths in the metadata JSONs will need to be fixed up. Or is this what the skip and limit parameters are intended for? Should I template the job file to start many jobs to deal with e.g. 10 videos at a time this way? If so I suppose these numbers include all files, or just mp4 files? If this is the recommended mechanism for HPC extractions, we should definitely document it.

lucaro:
How many feature modules are you using during the extraction? If you are only using a small number, the decoding and 'house-keeping' overhead might be larger than the extraction workload, so you might not get a high resource utilization independent of the queue size. Since cineast does not have the possibility to work on multiple videos simultaneously, starting multiple instances in parallel would be your easiest option here. You should be able to set the output directory for the JSON output in the job file, so avoiding conflicts when writing files should also be possible.

frankier:
I am just using the example configuration:

```
  "retriever": {
    "threadPoolSize": 2,
    "maxResults": 1000,
    "resultsPerModule": 1500,
    "features": {
      "semantic" : [
        {"feature": "ConceptMasksAde20k", "weight" : 1.0}
[...]

PR #118, opened 2020-07-24
Add relTo, allow file inputPath in extraction config

frankier:
This helps with usage in which the extraction process is invoked multiple times for each file within a larger collection -- typically by templating the extraction config file.


frankier:
This is related to https://github.com/vitrivr/cineast/issues/115 as well as the broader task of distributed extractions and extractions which run on computing clusters

PR #120, opened 2020-07-28
OpenApi Usage Fix and Javalin Update

sauterl:
Bumped Javalin to 3.9.1 and fixed the usage of the openapi plugin.

There are no breaking changes in this PR, but the usage of the [Javalin OpenAPI Plugin](https://javalin.io/plugins/openapi#getting-started) was fixed.

Also, a minor upgrade to the datetime parsing of json-annotated historical images is included in this PR, which is used in the context of the city-stories project.

**This is a mandatory PR before v3.0 is releasable**


sauterl:
TL;DR: We used openapi annotation based documentation, which is not supported in the way we setup the REST API. Hence, we now use a pseudo-DSL style documentation approach, which produces valid and useful OpenAPI Specs

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/StatusInvokationHandler.java
```diff
@@ -6,6 +6,8 @@
 import io.javalin.plugin.openapi.annotations.OpenApi;
```
silvanheller, line 6: While we're at it, this class should be called Statusinvo**c**ationhandler (to invoke something is an invocation with a c)

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/bool/FindDistinctElementsByColumnPostHandler.java
```diff
@@ -62,20 +54,34 @@ public DistinctElementsResult performPost(ColumnSpecification specification, Con
...
     return ROUTE;
   }
+  
+  
+  @Override
+  public OpenApiDocumentation docs() {
+    return OpenApiBuilder.document()
+        .operation(op -> {
+          op.summary("Find all distinct elements of a given column");
+          op.description("TODO");
```
silvanheller, line 80: I think for most handlers the description matches the summary so we can do the same thing here. One thing to note is that this handler caches results

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/segment/FindSegmentSimilarPostHandler.java
```diff
@@ -2,6 +2,8 @@
 
 import io.javalin.http.Context;
 import io.javalin.plugin.openapi.annotations.*;
+import io.javalin.plugin.openapi.dsl.OpenApiBuilder;
```
silvanheller, line 5: I think in this file the OpenAPI annotation for performPost() should also be removed?

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/session/ExtractItemHandler.java
```diff
@@ -2,6 +2,8 @@
 
 import io.javalin.http.Context;
 import io.javalin.plugin.openapi.annotations.*;
+import io.javalin.plugin.openapi.dsl.OpenApiBuilder;
```
silvanheller, line 5: Similar comment here: Why is the OpenApi annotation not removed?

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/session/StartExtractionHandler.java
```diff
@@ -2,6 +2,8 @@
 
 import io.javalin.http.Context;
 import io.javalin.plugin.openapi.annotations.*;
+import io.javalin.plugin.openapi.dsl.OpenApiBuilder;
```
silvanheller, line 5: OpenApi annotation is not gone

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/tag/FindTagsByIdsPostHandler.java
```diff
@@ -2,6 +2,8 @@
 
 import io.javalin.http.Context;
 import io.javalin.plugin.openapi.annotations.*;
+import io.javalin.plugin.openapi.dsl.OpenApiBuilder;
```
silvanheller, line 5: OpenApi annotation is not removed

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/tag/FindTagsGetHandler.java
```diff
@@ -2,6 +2,8 @@
 
 import io.javalin.http.Context;
 import io.javalin.plugin.openapi.annotations.*;
+import io.javalin.plugin.openapi.dsl.OpenApiBuilder;
```
silvanheller, line 5: OpenApi Annotation is not removed

citystories.json
```diff
@@ -0,0 +1,163 @@
+{
```
silvanheller, line 1: I don't think this file was meant to be included

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/bool/FindDistinctElementsByColumnPostHandler.java
sauterl: Thanks for the note, it's written in the new description.
I do think its important to have it there, as the description is where longer, well description, goes.
This is rather a hint for future handlers to include these and one might start an offensive to fully document the api...

citystories.json
sauterl: Was removed in the latest commit

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/handlers/actions/StatusInvokationHandler.java
sauterl: It appears that this typo was there since the very beginning - it was fixed now

sauterl:
> At some points, i'm unsure about why the OpenApi Annotation was not removed and the citystories config is probably not meant for dev

They simply were overlooked during the removal process. Thanks for pointing at those files were I missed the removal

PR #122, opened 2020-08-01
Merge new cottontail api => dev

frankier:
Would it be possible to merge this into dev? That way I can follow and make PR requests against the dev branches on cineast/vitrivr/cottontail concurrently and they remain compatible..


silvanheller:
Unfortunately, the dev branch includes (among other very useful things) also changes to the way results are sent which worsens performance for large resultsets. Therefore, this first needs to be reverted on the cottontail side before we can merge.

@ppanopticon can you revert the api change on the cottontail side?

silvanheller:
Closing since we're waiting for cottontail-db anyway.

PR #124, opened 2020-08-10
Extraction improvements

Spiess:
## Changes

- Implemented skipping of directory paths during extraction, since the directory files themselves are not processed by any extractors and only print unsupported file type debug messages
- Removed unused and deprecated fields in InputConfig
- Removed some empty Javadoc comments and minor cleanup

### DIFF
--- a/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/config/InputConfig.java
+++ b/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/config/InputConfig.java
@@ -32,7 +32,7 @@ public class InputConfig {
         this.path = path;
     }
 
-    @JsonProperty(required = true)
+    @JsonProperty()
     public String getRelTo() {
         return relTo;
     }
--- a/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/config/InputConfig.java
+++ b/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/config/InputConfig.java
@@ -13,8 +13,6 @@ public class InputConfig {
     private String name;
     private Integer depth = 1;
 
-    private Integer skip = 0;
-    private Integer limit = Integer.MAX_VALUE;
     private IdConfig id = new IdConfig();
 
     @JsonProperty
@@ -47,21 +45,5 @@ public class InputConfig {
     }
     public void setDepth(Integer depth) {
         this.depth = depth;
-    }
-
-    @JsonProperty
-    public Integer getSkip() {
-        return skip;
-    }
-    public void setSkip(Integer skip) {
-        this.skip = skip;
-    }
-
-    @JsonProperty
-    public Integer getLimit() {
-        return limit;
-    }
-    public void setLimit(Integer limit) {
-        this.limit = limit;
     }
 }
--- a/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/config/InputConfig.java
+++ b/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/config/InputConfig.java
@@ -11,6 +11,7 @@ import org.vitrivr.cineast.core.config.IdConfig;
 public class InputConfig {
     private String path;
     private String name;
+    private String relTo;
     private Integer depth = 1;
 
     private IdConfig id = new IdConfig();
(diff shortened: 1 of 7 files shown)


cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
```diff
@@ -363,6 +364,12 @@ private void shutdown() {
         continue;
       }
       ExtractionItemContainer item = providerResult.get();
+
+      /* Skip directories. They will be traversed recursively anyway, but obviously the actual directory is not processed */
+      if (Files.isDirectory(item.getPathForExtraction())) {
+        continue;
```
lucaro, line 370: Have you checked if this still works with image sequences, which are stored as a directory of (flat) images?

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
```diff
@@ -383,15 +388,17 @@ private void shutdown() {
       }
 
       /* Get the appropriate handler for this item. We ignore image sequences if they're not specified because they support the same file types as images*/
-      if (handlerCache.entrySet().stream().filter(handler -> handler != null && handler.getKey() != MediaType.IMAGE_SEQUENCE)
-          .anyMatch(handler -> handler.getValue() != null && handler.getValue().getKey() != null && handler.getValue().getKey().supportedFiles().contains(type))) {
-        return new ImmutablePair<>(item, handlerCache.entrySet().stream()
-            .filter(handler -> handler != null && handler.getValue().getKey() != null && handler.getValue().getKey().supportedFiles().contains(type) && handler.getKey() != MediaType.IMAGE_SEQUENCE).findFirst()
-            .get().getKey());
+      Optional<Map.Entry<MediaType, Pair<Decoder, Segmenter>>> match = handlerCache.entrySet().stream().filter(handler -> handler != null && handler.getValue().getKey() != null && handler.getValue().getKey().supportedFiles().contains(type) && handler.getKey() != MediaType.IMAGE_SEQUENCE).findFirst();
```
silvanheller, line 391: This is the same functionality, but removes code duplication

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
```diff
@@ -383,15 +388,17 @@ private void shutdown() {
...
+      if (match.isPresent()) {
+        return new ImmutablePair<>(item, match.get().getKey());
       } else {
-        /* If no appropriate handler is found, we log an error and continue with the next item */
-        LOGGER.error("No matching handlers found for type {} and item {}", type, item);
-        handlers.forEach((key, value) -> LOGGER.debug(key + " | " + value));
+        if (Files.isDirectory(item.getPathForExtraction())) {
+          LOGGER.debug("Not processing directory {}. Its content might be processed, but the directory itself is only processed for IMAGE_SEQUENCES. If you wish to use those, please specify the type explicitly in the extraction config.", item.getPathForExtraction());
```
silvanheller, line 396: This addresses the comment of @lucaro . Skipping directories was implemented because there was confusion about why folders were logged with an error-message

silvanheller:
This also includes PR #118 now. It updates some documentation & logging which caused confusion. Also, the documentation on the wiki of the extraction config is updated. To clarify also here: the `type` parameter never caused any problems, it is just deprecated for all non-LSC extraction. The reason there was confusion about its usage is that there were two getters of the `type` parameter in the `IngestConfig` and the second one was not detected as a getter by IntelliJ. After this PR, there is only one getter.

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/config/InputConfig.java
```diff
@@ -31,6 +32,14 @@ public void setPath(String path) {
         this.path = path;
     }
 
+    @JsonProperty(required = true)
+    public String getRelTo() {
```
lucaro, line 36: If such a property is required, could we at least call it something like 'basePath' or 'relativToPath' or something at least somewhat intuitive in the config?

silvanheller: Good catch, that should not be required.

Issue #121, opened 2020-07-28
Optimize should be run automatically or is misleadingly named


frankier:
Currently it appears that optimize must be run after an import for indices to be build in cottontail. This means it's required after an import for things to work correctly so is more than just an optimization step. I know this is all in a bit of a state of flux, but I'm just filing this issue (here, but perhaps it could equally well be filed against cottontail) to note that either:

 * Optimize should be run automatically at some point
 * *OR* optimize should be renamed to `index` (or even commit?) and it should be noted that it's compulsory to run it after adding any new data.

Note also that it's not just that queries won't work, but also integrity problems such as duplicate keys won't be reported until optimize is run (a reasonable design here I think but it should be made a bit clearer).

silvanheller:
What do you think about adding a `--finalize` flag to the `extract` command which is set to `true` by default? This way, the command is run after extraction by default, thereby eliminating the confusing behavior but if people are using external parallelism, they can disable the flag to have the same performance as now.

frankier:
Sounds like a good option to me!

Spiess:
@silvanheller the `doNotFinalize` flag you added is true by default, meaning optimization is not run. In fact, I was unable to disable the flag to trigger automatic optimization. Is this by design?

frankier:
Whoops! Another thing I realized after the PR was merged was that `optimize` should also possibly run by default after an `import`.

Personally, I don't necessarily mind an extra step (although in the common case it is nice to not have to do it separately) and it's really more about the naming.

PR #125, opened 2020-08-10
Automatically executing optimize after extraction if no flag is set

silvanheller:
See discussion in #121. If cottontail is not used, the `OptimizeEntitiesCommand` simply returns without doing anything.


silvanheller:
@frankier can you have a look?

frankier:
Looks good to me!

PR #132, opened 2020-08-13
Extraction blocking

Spiess:
Implemented blocking in the extraction command.

The extraction command was previously run independently on a separate thread, which allowed other commands to be run from the CLI in parallel, potentially causing problems. This commit introduces a blocking call so that the main thread responsible for the CLI waits to join the extraction thread.

### DIFF
--- a/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/ExtractionCommand.java
+++ b/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/ExtractionCommand.java
@@ -53,8 +53,8 @@ public class ExtractionCommand implements Runnable {
               }
             });
           }
-          dispatcher.start();
           dispatcher.registerListener((ExtractionCompleteListener) provider);
+          dispatcher.start();
           dispatcher.block();
         } else {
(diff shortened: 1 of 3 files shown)


silvanheller:
@lucaro thoughts? My intuition is that the CLI is specifically supposed to be non-blocking, but i can see the argument that it is also supposed to only return when the actual CLI call is done.

lucaro:
Yes, having the CLI active was the idea, in order to be able to interact with the stack even when it is busy. In practice, this is however usually not very helpful as the log writes to the same stream as the CLI, but that's an independent problem...

silvanheller:
Upon further discussion, both the `import` and `setup` command are blocking. So I'd argue that to avoid confusing the user (and accidents when executing stuff in parallel), we make the extraction command blocking as well. If users want to execute additional commands, they can launch a second CLI or simply launch the extraction-command via jar.

lucaro:
Hmm, can we at least add a possibility to interrupt the blocking commands? That would probably also be useful for the import process, in case one notices something being wrong half way through.

silvanheller:
If something goes wrong during the import process, exiting the cli is anyway necessary probably since there need to be changes to multiple things. We're also not sure how an abort would be structured since Cineast currently doesn't do transactions. I'm therefore merging this PR since it is definitely an improvement over the current behavior.

PR #133, opened 2020-08-13
Add dockerfile

frankier:
I have been using my own Singularity recipe up until now but:

 1. Building Docker images on Docker Hub for use with Singularity seems to actually be a better option in many cases than using Singularity + SingularityHub (more info: https://github.com/frankier/gsoc2020/wiki/Should-you-avoid-using-Singularity-and-Singularity-Hub-to-build-your-Singularity-images%3F )
 2. Docker is commonplace so this is presumably more useful to others
 3. Docker Hub makes it easy for everyone to get autobuilds of their own branches/PRs

This exposes a custom entrypoint so we can choose to use either `docker run cineast api` or `docker run cineast cli` to run either the server+repl or the cli.

Ready to merge if the Docker Hub build works https://hub.docker.com/repository/registry-1.docker.io/frankierr/cineast/builds/5cf8cad5-0671-47ca-b2bd-6d3d96a55888

See also: https://github.com/vitrivr/cottontaildb/pull/44

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,43 @@
+FROM openjdk:14-buster AS build
+
+RUN apt-get update && \
+  apt-get install -y maven
+COPY . /cineast-src
+RUN cd /cineast-src && \
+  ./gradlew :cineast-core:generateProto && \
+  ./gradlew getExternalFiles && \
+  ./gradlew fatJar
+
+FROM openjdk:14-slim-buster
+
+RUN mkdir -p /opt/cineast
+
+COPY --from=build \
+  /cineast-src/cineast.json \
+  /opt/cineast/cineast.json
+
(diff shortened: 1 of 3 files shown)


silvanheller:
Sounds like a very good idea to me, we'll do some testing on our side and see if we have other ideas / feature requests.

frankier:
This is ready to merge now. You can test with e.g. `docker run frankierr/cineast cli cineast.json help`.

Spiess:
I have built and pushed the docker image to vitrivr/cineast. Ideally we would also add a [CI hook and workflow](https://docs.github.com/en/actions/language-and-framework-guides/publishing-docker-images) to automatically build and push a new docker image on each new Cineast release, but that's a new and separate task.

PR #136, opened 2020-08-14
Updated thumbnail exporters.

Spiess:
Changes:
- Set default thumbnail image type to PNG
- Modified thumbnail exporters only to write thumbnail images if the filetype is actually supported by the exporter
- Modified thumbnail exporters only to create mediaobject thumbnail directory if the filetype is actually supported and is able to be processed by the exporter


Spiess:
> Thumbnails were set to JPG on purpose, since lossless compression offers no benefit in this case but increases the output file size substantially.

Ok, this does make sense and I will change it back to JPG. However the ShotThumbnailsExporter was already defaulting to PNG, should I also change this to JPG or should the additional quality be preserved for video/image thumbnails?

lucaro:
This setting is probably a leftover from an older mechanism, setting the exporter to JPG by default appears to be the most sensible option to me.

Spiess:
I have now changed the default file type for thumbnails to JPG as well as fixed a previously unidentified problem which would prevent thumbnail files with alpha channels from being written to JPG (even if there was no transparency in the image).

cineast-core/src/main/java/org/vitrivr/cineast/core/features/exporter/AudioSpectogramExporter.java
```diff
@@ -91,7 +96,7 @@ public void processSegment(SegmentContainer shot) {
             BufferedImage image = AudioSignalVisualizer.visualizeSpectogram(spectrums, this.width, this.height);
             if (image != null) {
                 Files.createDirectories(directory);
-                ImageIO.write(image, "JPEG", directory.resolve(shot.getId() + ".jpg").toFile());
+                ImageIO.write(image, format, directory.resolve(shot.getId() + format.toLowerCase()).toFile());
```
lucaro, line 99: There's a `.` missing in the file name

PR #137, opened 2020-08-14
Update Conttontaildb-proto

frankier:


silvanheller:
Is this compatible with `master` in cottontail? Otherwise i'd wait for @ppanopticon to merge `dev` into `master` in cottontail and then update (Sidenote: I'm generally unhappy with the current build setup, it would be more ideal to publish jars...)

frankier:
Okay this is a bit of a confusing one for me. I have my main working branch from GSOC developed against vitrivr dev + cineast dev and then two PRs against cottontaildb-proto and cottontaildb, both against the dev branch. But now I see that cineast dev is only compatible with the master branch of cottontaildb/cottontaildb-proto. So am I to understand that I should rebase my cottontaildb PRs against master? It seems a bit odd to develop simultaneously against master and dev and seems like it's likely to cause confusing merges and confusion about canonical versions but if this is how your team works, I'm willing to try to adapt.

Could this situation either be simplified, or could someone perhaps give me some direction here? @lucaro

lucaro:
From what I understand, and @silvanheller or @ppanopticon correct me if I'm wrong, is that cottontail was not supposed to have a master/dev split like cineast or vitrivr-ng have just yet, since it has not reached a 1.0 state yet anyway and so master is not considered to be 'stable' even though it pretty much is already. The dev branch was introduced there to change some things which were on master first but had unanticipated consequences, especially in combination with cineast. So I guess the vitrivr stack is using cottontail at master for the moment since dev is more experimental. I guess we need some consistent naming conventions...

frankier:
Thank you for clarifying. Going forward I'll ignore the dev branch on cottontaildb so all my PRs there have been rebased against master.

silvanheller:
@lucaro is correct. The confusion is understandable and one more reason to switch to jar publishing for cottontail-proto and version-numbering for cottontail. I'll close this PR and we'll wait for a new `master` branch in cottontail (which will probably happen end of august)

PR #138, opened 2020-08-14
Import command safety

Spiess:
Running import without type or input arguments would cause a NullPointerException and crash the CLI.

Fixed to only print a descriptive error message and return to the CLI.


cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/ImportCommand.java
```diff
@@ -51,6 +51,14 @@
 
   @Override
   public void run() {
+    if (type == null) {
```
silvanheller, line 54: Use `@Required` instead

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/ExtractionCommand.java
```diff
@@ -23,7 +23,7 @@
 public class ExtractionCommand implements Runnable {
 
   @Option(name = {"-e", "--extraction"}, title = "Extraction config", description = "Path that points to a valid Cineast extraction config file.")
-  private String extractionConfig;
+  private String extractionConfig = null;
```
silvanheller, line 26: Also `@Required` here, then you can remove both initialization and null-checks

PR #140, opened 2020-08-24
Pose querying, red hen importers and skeleton extraction service

frankier:
Features added:
 * Querying and importing of literal pose embeddings
 * Querying and importing of semantic hand embeddings
 * Importing of Red Hen/NewsScape subtitle and OCR data
 * Skeleton extraction service for Vitrivr -- skeletons are sent back to the client so the user can inspect them during query construction


frankier:
To merge this:

 * https://github.com/vitrivr/cottontaildb-proto/pull/7 needs to be merged and the submodule updated;
 * You might like to mirror the resources and modify the resources CSV.

Please let me know if there are any problems with this PR.

silvanheller:
Thanks for all the work! I’ll close this PR since there is too much divergence between the fork and the current codebase and we’re lacking the resources to integrate & review this. If the features implemented are used in future projects, we’ll happily help to integrate the code in the codebase in a new PR.

Issue #89, opened 2019-09-02
Release Management Automation


lucaro:
Since the last release, nobody updated the pre-built version of the UI in the github releases. Could someone have a look at this and how to prevent this in the future?

silvanheller:
What should the release artifact include? just jar and config-file? Or the entire source code? One way to prevent this is to have a PR template.

lucaro:
Ideally, there would be a pre-built version for every component of the stack which is known to be compatible to each other. This is especially relevant for the UI (see https://github.com/vitrivr/vitrivr-ng/issues/15 where I copied this from) since having a jdk when you have a jre is probably more likely than having a node environment. We could then link all these released components together on a page on vitrivr.org.

silvanheller:
The current state is that we'd like two things: 
- Automatically update `latest` on Dockerhub (See [Github Actions](https://docs.github.com/en/actions/language-and-framework-guides/publishing-docker-images)
- Automatically build & publish releases (always / for certain versions)

silvanheller:
Thanks @mpas97!

PR #142, opened 2020-09-22
build release and push docker image to hub

mpas97:
Being part of #89, this generates a latest & _version_ tagged build to docker hub.


mpas97:
@silvanheller After merging, you can recreate the v3 tag then it should build a 3.0.0 and latest tag on docker hub.

mpas97:
Finally, the jars are now also added to the release assets too.

PR #182, opened 2021-03-15
New Cottontail DB API

ppanopticon:
This PR reflects the latest changes and additions to the Cottontail DB API. Most importantly, the DB-Facade implementations now use the new _SimpleClient_. Requires Cottontail DB version _0.12.0 or higher_.

I tested the basic functionality (NNS & text queries, setup, data import) but I guess it makes sense that others try this as well.

**Important:** Changes to the catalogue introduced in version 0.12.0 render old deployments of Cottontail DB incompatible with the new version. There is a migration CLI command `system migration --input /path/to/config`, which takes care of upgrading old deployments to new ones. For more details on migration, see the [Cottontail Wiki](https://github.com/vitrivr/cottontaildb/wiki/Catalogue-Migration)


lucaro:
I'll certainly have a look, I do however find the "All checks have failed" already somewhat discouraging...

ppanopticon:
> I'll certainly have a look, I do however find the "All checks have failed" already somewhat discouraging...

I share the feeling. :-) But the failing tests are integration tests that require a running Cottontail DB instance, which is why they are failing. I'm not sure if it is even possible for them to work. I'll look into it.

silvanheller:
Few preliminary thoughts:
- The Database-tests should be skipped if ping() fails. The new `SimpleClient` caused the `ping()` implementation of the `CottontailSelector` to behave wrongly, which I've fixed in the latest commit.
- Now that cottontail runs dockerized, it should be possible to have the cottontail docker-container as a prerequisite - this would allow us to run all tests. (Currently, we would not even notice if the changes would break)

I'll have a look at the actual changes, but this should stop the CI from failing.

cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailWrapper.java
```diff
@@ -1,339 +1,54 @@
...
+  public final ManagedChannel channel;
 
-      @Override
-      public void onError(Throwable t) {
-        status[0] = true;
-        status[1] = true;
-        LOGGER.error("Error during insert. Everything was rolled back: {}", t.getMessage());
-      }
+  /** The {@link SimpleClient} instance that facilitates access to Cottontail DB. */
+  public final SimpleClient client;
```
silvanheller, line 22: On `dev`, simply running `ping()` from Cineast on windows causes the following error Client-Side: `CANCELLED: RST_STREAM closed stream. HTTP/2 error code: CANCEL`

The cottontail-output is as follows: `Mar 15, 2021 9:32:59 AM io.grpc.internal.ServerCallImpl internalClose
WARNING: Cancelling the stream with status Status{code=INTERNAL, description=Completed without a response, cause=null}`.

I've fixed this in a1baab5e3ce2efcf758d7d7e2e6f9b46445edd34 on the cottontail `dev` branch

silvanheller (changes requested):
Now that `ping()` is working, tests can be executed and there are a lot of errors. @ppanopticon can you have a look at how the tests should use the new api for entity-creation etc.?`There is an `org/vitrivr/cineast/core/db/cottontaildb/CottontailIntegrationTest.java` which should test a lot of relevant features.

ppanopticon:
All integration tests should be running now: They helped a great deal in detecting and fixing a few Cottontail DB bugs. 

I'll now check whether we can use a Cottontail DB Docker container to run tests in CI.

silvanheller (commented):
Thanks! I'll have a look sometime this week if migration & the new api works well on the vbs data

silvanheller (changes requested):
I've added new Unit tests (and refactored a bit), the `CottontailBooleanIntegrationTest` being the most relevant. Currently, both IN() and BETWEEN() clauses do not work due to the way the `SimpleClient` casts variables. I have no strong opinions on how to solve this problem.

ppanopticon:
> I've added new Unit tests (and refactored a bit), the `CottontailBooleanIntegrationTest` being the most relevant. Currently, both IN() and BETWEEN() clauses do not work due to the way the `SimpleClient` casts variables. I have no strong opinions on how to solve this problem.

Yes there is indeed a problem here, which can be fixed though. However, even with the fix, there the test still fails because it doesn't yield any results. I quickly checked and to me it seems that you're querying the wrong column.

On line 182 and 195, wouldn't you rather query DATA_COL_NAME_1 instead of DATA_COL_NAME_3 (the latter does not hold any values in the specified range).

silvanheller:
> Yes there is indeed a problem here, which can be fixed though. However, even with the fix, the test still fails because the query doesn't yield any results. I quickly checked and to me it seems that you're querying the wrong column.
> 
> On line 182 and 195, wouldn't you rather query DATA_COL_NAME_1 instead of DATA_COL_NAME_3? The latter does not hold any values in the specified range.

Thanks, I fixed the unittests

ppanopticon:
Perfect! I just pushed a fix for the two issues with BETWEEN and IN. This should work now as well.

On my setup, however, I have four failing tests:

- "Text: Three els, one of those with quotes"
- "Text: One el (Two words), Fuzzy"
- "Text: One el (two words), quotes"
- "Text: One el, no quotes"

However, I'm currently running on an experimental version and I remember these tests running successful on master. Do they work for you?

ppanopticon:
> Perfect! I just pushed a fix for the two issues with BETWEEN and IN. This should work now as well.
> 
> On my setup, however, I have four failing tests:
> 
> * "Text: Three els, one of those with quotes"
> * "Text: One el (Two words), Fuzzy"
> * "Text: One el (two words), quotes"
> * "Text: One el, no quotes"
> 
> There is a slight mismatch in the returned number of results. However, I'm currently running on an experimental version and I remember these tests running successful on master. Do they work for you?

Never mind! It works now - it was indeed a problem with the branch which should not occur on master

silvanheller (changes requested):
I just re-enabled a basic text retrieval test in the `BooleanIntegrationTest` which causes a recursion error at UnaryLogicalOperatorNode.copyWithGroupInputs(UnaryLogicalOperatorNode.kt:71) in Cottontail

ppanopticon:
This took a bit longer than expected but it should be fixed now. 

The unit test still fails though, I believe you're checking the wrong column - is this possible?

silvanheller:
The problem was that the current Cineast-Implementation of Fulltext Retrieval did not return all columns associated with a table. This should be fixed now.

silvanheller (approved):
Looks good! I've tested schema migration on both the VBS and LSC dataset and it looks like everything works well. All Unittests in Cineast run (and they cover a lot of things in the meantime, which i'm very glad about).

From my point of view we can merge, but maybe @lucaro has some thoughts on the API.

PR #197, opened 2021-04-02
 DBBooleanIntegrationTest Expand

### LINKED ISSUE #192: Expand DBBooleanIntegrationTest
The current `org.vitrivr.cineast.core.db.DBBooleanIntegrationTest` only tests a subset of possible invocations of the `DBSelector` class. Possible extensions include:
- Testing the `getRowsAND` method with different conditions
- More tests for the different `RelationalOperator` values

Note: Currently only available on the `new-cottontail-api` branch, see #182

Liuhaai:
I added four tests into current DBBooleanIntegrationTest.java
1. getRows for Greater() test
2. getRows for Less() test
3. getRowsAND for BETWEEN() AND BETWEEN() query
4. getRowsAND for BETWEEN() AND IN() query
ALL tests in DBBooleanIntegrationTest pass the JUnit tests. Might Closes #192


silvanheller:
Thanks for the PR! I'll have a look, but I don't think we'll close #192 since it's a good issue for people to get started and there's probably an infinite number of tests you could implement

silvanheller:
Migrating to #206

PR #201, opened 2021-05-07
Api documentation and checkstyle enforcment

vGsteiger:
In order to understand the current code better and to get familiar with the developments over time, I spent some time improving maintainability by updating the documentation and enforcing a common checkstyle at various locations of the API package.

I did not alter any logic and only applied reformatting rules based on the file in the `docs/` folder.

I tried to develop a consistent docstyle over all classes from existing documentation already written.

- Enforced the current vitrivr checkstyle on various classes in the API package
- Extended and wrote additional documentation on various classes in the API package


silvanheller (changes requested):
Thanks for the work! In general, i dislike @version tags because they implicitly require changelogs between file versions.

vGsteiger:
@silvanheller 
> Thanks for the work! In general, i dislike @Version tags because they implicitly require changelogs between file versions.

Thanks for the feedback, I removed the @Version tags from the documentation.

PR #204, opened 2021-05-15
Standardize toString in API messages

vGsteiger:
Applied the same toString method to all API messages to have a unified form of logging:

- Applied the in #203 discussed style to all API message classes
- Some minor code refactoring (spaces and import orders)

The toString method employed was:
`return ReflectionToStringBuilder.toString(this, ToStringStyle.JSON_STYLE);`


silvanheller:
Thanks!

PR #206, opened 2021-05-18
New Boolean Integration Tests (originally #197)

silvanheller:
PR by @Liuhaai, who added tests. Partially addresses #192.


silvanheller:
@ppanopticon, one test currently fails and it looks like to me like this is a cottontail-bug. Specifically, BETWEEN(10,11) AND BETWEEN(11,12) returns (10,11,12) instead of just 11.
Can you have a look?

ppanopticon:
> @ppanopticon, one test currently fails and it looks like to me like this is a cottontail-bug. Specifically, BETWEEN(10,11) AND BETWEEN(11,12) returns (10,11,12) instead of just 11.
> Can you have a look?

I cannot reproduce this issue in Cottontail DB 0.12.9. Could it be, that this has been resolved in the meanwhile?

PR #211, opened 2021-06-26
Support for IIIF Image API 2.1.1 and 3.0 servers

singaltanmay:
# Features
* Helps build image api URLs using Java method calls
* Prevents user from making unsupported request by throwing errors if feature not supported by the particular server
* Download image and the image information JSON onto the file system
* Integrated with the ExtractionCommand
* Simple JSON configuration using the `extract -e example_iiif_job.json` command
* Support for global and image level parameters such as size, rotation, quality etc
* Suports IIIF Image API 2.1.1 and 3.0
* Unit tests included


cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageRequest.java
```diff
@@ -0,0 +1,147 @@
...
+    this.size = size;
+    this.rotation = rotation;
+    this.quality = quality;
+    this.extension = extension;
+  }
+
+  /**
+   * Percent encodes "/","?","#","[","]","@" and "%" to their corresponding ASCII reserved characters
+   */
+  public static String percentEncode(String toEncode) {
```
lucaro, line 45: If I understand correctly, what you are attempting to do here is URL encoding. There are many more special characters that would need to be covered in this case. Luckily, there is already built-in logic for this, see java.net.URLEncoder.

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageRequest.java
```diff
@@ -0,0 +1,147 @@
...
+        + "."
+        + percentEncode(extension);
+  }
+
+  public void saveToFile(String filePath, String fileName) throws IOException {
+    URL url = new URL(this.generateIIIFRequestUrl());
+    BufferedImage img = ImageIO.read(url);
+    File file = new File(filePath + "/" + fileName + "." + this.getExtension());
+    ImageIO.write(img, this.getExtension(), file);
+    LOGGER.debug("Image downloaded and written to file successfully. Image url:\t" + url);
```
lucaro, line 140: In the logger, you can use '{}' as a placeholder for string interpolation, so the canonical way of doing this would be
```java
LOGGER.debug("Image downloaded and written to file successfully. Image url:\t{}", url);
```

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageRequestFactory.java
```diff
@@ -0,0 +1,404 @@
...
+        String url = iiifConfig.getBaseUrl() + "/" + identifier;
+        try {
+          response = ImageInformationRequest.fetchImageInformation(url);
+        } catch (IOException e) {
+          LOGGER.error("An error occurred while making an Image Information request with the url: " + url);
+          e.printStackTrace();
+        }
+        if (response != null) {
+          try {
+            new ObjectMapper().readValue(response, ImageInformation_v3.class);
```
lucaro, line 99: Isn't there a way to look at the response directly, rather than using a deserialization exception?

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/v2/ImageApiCompliance_v2.java
```diff
@@ -0,0 +1,88 @@
...
+      return true;
+    }
+    if (format.equals(EXTENSION_PNG)) {
+      return level.equals(LEVEL_2);
+    }
+    return false;
+  }
+
+  public static boolean isHttpFeatureSupported(String httpFeature, String level) {
+    switch (httpFeature) {
```
lucaro, line 80: Why use a switch if there is only one case?

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/v2/ImageInformation_v2.java
```diff
@@ -0,0 +1,431 @@
...
+ * @created 29.05.21
+ */
+public class ImageInformation_v2 implements ImageInformation {
+
+  private static final Logger LOGGER = LogManager.getLogger();
+  /**
+   * A list of profiles, indicated by either a URI or an object describing the features supported. The first entry in the list must be a compliance level URI.
+   */
+  @JsonProperty(required = true)
+  private List<Object> profile;
```
lucaro, line 29: Is `Object` really the most specific you can be here?

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/ExtractionCommand.java
```diff
@@ -39,6 +59,21 @@ public void run() {
...
         final JacksonJsonProvider reader = new JacksonJsonProvider();
         final IngestConfig context = reader.toObject(file, IngestConfig.class);
+        if (context != null) {
+          InputConfig inputConfig = context.getInput();
+          IIIFConfig iiifConfig = inputConfig.getIiif();
+          if (iiifConfig != null) {
+            String directoryPath;
+            String iiifConfigPath = inputConfig.getPath();
+            if (iiifConfigPath == null || iiifConfigPath.isEmpty()) {
+              directoryPath = file.getAbsoluteFile().toPath().getParent() + "/iiif-media";
```
lucaro, line 69: The target should probably be configurable, rather than being hard-coded to 'iiif-media'.

lucaro (commented):
Looks good so far. There are some minor things I don't quite understand, though.

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageApiVersion.java
```diff
@@ -0,0 +1,107 @@
...
+      throw new IllegalArgumentException(imageLevelString + " is not a supported IIIF Image API Compliance level");
+    }
+    return new ImageApiVersion(imageApiVersionEnum);
+  }
+
+  /**
+   * Converts numeric version number string into an enum item
+   */
+  private static IMAGE_API_VERSION numericStringToEnum(String input) {
+    if (input.equals(IMAGE_API_VERSION_2_1_1_NUMERIC)) {
```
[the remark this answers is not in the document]
singaltanmay: Done

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageApiVersion.java
```diff
@@ -0,0 +1,107 @@
...
+      return IMAGE_API_VERSION.THREE_POINT_ZERO;
+    }
+    return null;
+  }
+
+  /**
+   * Converts IIIF Image API level string into an enum item
+   */
+  private static IMAGE_API_VERSION apiComplianceLevelToEnum(String input) {
+    if (input.equals(IMAGE_API_VERSION_2_1_1_COMPLIANCE_LEVEL_2)) {
```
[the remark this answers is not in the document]
singaltanmay: Done

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageRequest.java
singaltanmay: Done

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/v2/ImageApiCompliance_v2.java
singaltanmay: Replaced with an `if` statement

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageInformation.java
```diff
@@ -0,0 +1,231 @@
...
+  /**
+   * Get the max area supported by server
+   */
+  Long getMaxArea();
+
+
+  /**
+   * Inner class used to parse the Image Information JSON response.
+   */
+  class SizesItem {
```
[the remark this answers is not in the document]
singaltanmay: This inner class is only used to serialize and deserialize members of the ImageInformation class and thus can be declared here instead of cluttering up the project tree

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageInformation.java
```diff
@@ -0,0 +1,231 @@
...
+    public SizesItem(long width, long height) {
+      this.width = width;
+      this.height = height;
+    }
+
+    public SizesItem() {
+    }
+
+    @java.lang.Override
+    public java.lang.String toString() {
```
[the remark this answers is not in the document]
singaltanmay: These annotations got autogenerated (by [delombok](https://projectlombok.org/features/delombok) script) in this format and have no special meaning and can be replaced by their shorted versions `@java.lang.Override -> @Override`

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageRequestFactory.java
singaltanmay: Deserilaization is a computationally cheap operation to perform and is only executed in the rare cases where the Image Api version is unknown. An alternative way would be to manually search for key-value substrings in the JSON but this method is prone to errors as there can be matching strings in the JSON that do not correspond to the data we are looking for and thus give a false negative.

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/v2/ImageInformation_v2.java
singaltanmay: `profile` key in version `2.1.1` and below can contain multiple different types of values which are not supported in Java and thus needs to be manually checked for type.

cineast-core/build.gradle
```diff
@@ -112,6 +112,9 @@ dependencies {
     api group: "com.google.guava", name: "guava", version: version_guava
     api group: "com.googlecode.javaewah", name: "JavaEWAH", version: version_javaewah
 
+    /** Mockito for unit testing */
+    implementation 'org.mockito:mockito-inline:3.11.0'
```
silvanheller, line 116: Why not mockito-core? https://stackoverflow.com/questions/65986197/difference-between-mockito-core-vs-mockito-inline

silvanheller, line 116: Why not mockito-core? https://stackoverflow.com/questions/65986197/difference-between-mockito-core-vs-mockito-inline

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/IIIFConfig.java
```diff
@@ -0,0 +1,190 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import org.vitrivr.cineast.core.iiif.imageapi.ImageApiVersion;
+
+/**
+ * IIIF configuration used to fetch media files from remote servers.
+ *
+ * @author singaltanmay
+ * @version 1.0
+ * @created 28.05.21
```
silvanheller, line 12: Please avoid the usage of @version and @created tags in javadoc, i know we have them sometimes in the codebase, but we're trying to get rid of them :)

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/IIIFConfig.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  @JsonProperty
+  public void setImageApiVersion(final String imageApiVersion) {
+    this.imageApiVersion = imageApiVersion;
+  }
+
+  @Override
+  public java.lang.String toString() {
+    return "IIIFConfig(imageApiVersion=" + this.getImageApiVersion() + ", baseUrl=" + this.getBaseUrl() + ", region=" + this.getRegion() + ", size=" + this.getSize() + ", rotation=" + this.getRotation() + ", quality=" + this.getQuality() + ", format=" + this.getFormat() + ", iiifItems=" + this.getIiifItems() + ")";
```
silvanheller, line 47: Use ` return ReflectionToStringBuilder.toString(this, ToStringStyle.JSON_STYLE);` - this is something we're trying to standardize (see e.g. https://github.com/vitrivr/cineast/pull/204)

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/IIIFConfig.java
```diff
@@ -0,0 +1,190 @@
...
+  public List<IIIFItem> getIiifItems() {
+    return this.iiifItems;
+  }
+
+  @JsonProperty("items")
+  public void setIiifItems(final List<IIIFItem> iiifItems) {
+    this.iiifItems = iiifItems;
+  }
+
+  public static class IIIFItem {
```
silvanheller, line 113: Please create a separate class file for this

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/IIIFConfig.java
```diff
@@ -0,0 +1,190 @@
...
+    @JsonProperty
+    private String quality;
+    @JsonProperty
+    private String format;
+
+    public IIIFItem() {
+    }
+
+    @Override
+    public java.lang.String toString() {
```
silvanheller, line 132: Same comment for toString() methods applies here

example_iiif_job.json
```diff
@@ -0,0 +1,95 @@
...
+    "skip": 0,
+    "id": {
+      "name": "SequentialObjectIdGenerator",
+      "properties": {}
+    },
+    "iiif": {
+      "url": "https://libimages.princeton.edu/loris/pudl0001/5138415",
+      "imageApiVersion": "2.1.1",
+      "rotation": 10,
+      "items": [
```
silvanheller, line 14: If i understand this example correctly, IIIF does not have a discovery mechanism, correct?

silvanheller (changes requested):
Looks very interesting, thanks for your work! I'll let @Spiess and @lucaro review the IIIF-code since that is not my area of expertise, but i've added some general comments on codestyle. Regarding the @author / @created / @version tags, i've noticed that we only unified the javadoc style in the api project, i just did so in the core project as well.

cineast-core/build.gradle
singaltanmay: There are a few unit tests where I am partially mocking some classes and mockito-core does not support partial mocks

example_iiif_job.json
singaltanmay: Image API does'nt have a dicovery mechanism, IIIF has a "Change Discovery" API but it is currently in Beta and no implementing servers are available for us to make use of. Currently, I am working on integrating the Presentation API which let's us download all images in a manifest at once. For example, an entire book can be downloaded using a single URL specified in the job file. You can check out code this at https://github.com/singaltanmay/cineast/tree/feature/IIIF_Presentation_API

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/IIIFConfig.java
singaltanmay: I have removed @author @version and @created tags from all the Javadoc I have created

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/IIIFConfig.java
singaltanmay: Done

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/IIIFConfig.java
singaltanmay: Done

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/IIIFConfig.java
singaltanmay: Done

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/ExtractionCommand.java
singaltanmay: Contrary to what I said in the meeting yesterday, this is a feature not a bug. The `/iiif-media` path is a fallback which used only when the user has not specified a directory in the config file.

cineast-core/src/main/java/org/vitrivr/cineast/core/iiif/imageapi/ImageRequest.java
singaltanmay: Replaced this method URLEncoder

PR #212, opened 2021-07-01
OCRSearch feature implementation

alexandertheus:
Implementation of the OCRSearch feature extractor for still images and videos.

Scene Text Detection: EAST
Scene Text Recognition: VGG-BiLSTM-CTC
Scene Text Extraction for Video: HyText (explained in bachelor thesis)

Scene Text Recognition module can be used for any CTC-based recognition (e.g. CRNN)
HyText allows for many custom modifications. The most notable one being the rate, which determines the rate at which detections are done. A higher rate leads to increased inference times, but slightly lower accuracy. Currently, one can choose between BoofCV's Circulant and TLD tracker (default is set to Circulant). The Circulant tracker should be chosen if high inference time is of importance.


cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractTextRetriever.java
```diff
@@ -80,11 +80,6 @@ public void init(PersistencyWriterSupplier phandlerSupply, int batchSize) {
     this.writer = new SimpleFulltextFeatureDescriptorWriter(phandlerSupply.get(), this.tableName, batchSize);
   }
 
-  @Override
-  public void processSegment(SegmentContainer shot) {
```
silvanheller, line 84: I don't see why you would remove this and then re-add it twice in implementing classes, especially since there are no plans to add this for Description, TagsFt and AudioTranscription

silvanheller, line 84: I don't see why you would remove this and then re-add it twice in implementing classes, especially since there are no plans to add this for Description, TagsFt and AudioTranscription

cineast-core/src/main/java/org/vitrivr/cineast/core/util/HungarianAlgorithm.java
```diff
@@ -0,0 +1,337 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/**
+ * An implementation of the Hungarian algorithm for solving the assignment
```
silvanheller, line 27: Ideally, you would add a comment to this class indicating its purpose in the context of the codebase

cineast-core/src/main/java/org/vitrivr/cineast/core/util/MultiTracker.java
```diff
@@ -0,0 +1,82 @@
...
+import boofcv.abst.tracker.TrackerObjectQuad;
+import boofcv.factory.tracker.FactoryTrackerObjectQuad;
+import boofcv.struct.image.GrayU8;
+import georegression.struct.shapes.Quadrilateral_F64;
+import org.vitrivr.cineast.core.data.Pair;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class MultiTracker {
```
silvanheller, line 12: Same comment as in the `HungarianAlgorithm`, documentation putting this into context would be appreciated

cineast-core/src/main/java/org/vitrivr/cineast/core/features/OCRSearch.java
```diff
@@ -1,24 +1,510 @@
...
+   * @return The transformed string
+   */
+  private String transformString(String str) { return str.replace("i", "l").replace("l", "1").replace("a", "o");}
+
+  /**
+   * safeText transforms and safes the scene text in the database
+   * @param id id of the shot (the {@link SegmentContainer}) to be processed
+   * @param recognition The scene text that should be safed
+   */
+  private void safeText(String id,String recognition) {
```
silvanheller, line 202: I think this should be `save`

cineast-core/src/main/java/org/vitrivr/cineast/core/features/OCRSearch.java
```diff
@@ -1,24 +1,510 @@
...
+
+    return spatial_dis + inv_IoU + 10*edit_dis + 0.2*Math.abs(streamFirst.getFirst() - streamLast.getLast());
+  }
+
+  /**
+   * transformString transforms the string by replacing certain characters with others
+   * @param str The string to be transformed
+   * @return The transformed string
+   */
+  private String transformString(String str) { return str.replace("i", "l").replace("l", "1").replace("a", "o");}
```
silvanheller, line 195: Why would you replace those characters? Are there known bugs in the method?

cineast-core/src/main/java/org/vitrivr/cineast/core/features/OCRSearch.java
```diff
@@ -1,24 +1,510 @@
...
+    return count==0 ? 0 : total_IoU / count;
+  }
+
+  /**
+   * Extracts the scene text and ingests it using the {@link SimpleFulltextFeatureDescriptor}.
+   * @param shot The {@link SegmentContainer} that should be processed.
+   */
+  @Override
+  public void processSegment(SegmentContainer shot) {
+    TextDetector_EAST detector = new TextDetector_EAST().initialize();
```
silvanheller, line 258: Can detector / recognizer be re-used? If so, it would be very nice to make them thread-safe and reuse them. Otherwise new instances are created per shot, which is costly for large data sets.

silvanheller (commented):
@lucaro If this should be merged, we should move the large files in the resource-folder to data.vitrivr.org so they can be fetched with `./gradlew getExternalFiles`.

@alexandertheus this looks very promising, thanks for the work! I've left some minor comments. Please use the java code style in the repo (https://github.com/vitrivr/cineast/blob/master/docs/intellij-java-google-style.xml) to reformat your new code, from a first glance it looks to me like it isn't formatted yet.

lucaro:
@silvanheller yes, we would absolutely move the large files out of the repository afterwards, it was just easier to include them in the PR so they don't get lost.

cineast-core/src/main/java/org/vitrivr/cineast/core/features/OCRSearch.java
lucaro: The way I understand this is that this is a simple way to introduce an invariance against some somewhat commonly confused characters.

silvanheller: That makes total sense, in this case i guess some documentation would be helpful @alexandertheus

alexandertheus: Will do! Upper case "i" and lowercase "l", and 1 are almost visually indistinguishable and the text recognizer hence has a hard time distinguishing them. Same can be said about "a" and "o", although to a lesser extent.

alexandertheus:
@silvanheller Thanks for having a look at my code! I will go through your comments and make changes accordingly.

cineast-core/src/main/java/org/vitrivr/cineast/core/util/MultiTracker.java
alexandertheus: Is resolved in the new commit. I additionally made clarifying comments in other classes I created.

cineast-core/src/main/java/org/vitrivr/cineast/core/util/TextDetector_EAST.java
```diff
@@ -0,0 +1,163 @@
...
+                    double sinA = Math.sin(angle);
+                    double x0 = x0Data.get(0, x)[0];
+                    double x1 = x1Data.get(0, x)[0];
+                    double x2 = x2Data.get(0, x)[0];
+                    double x3 = x3Data.get(0, x)[0];
+                    double h = x0 + x2;
+                    double w = x1 + x3;
+                    Point offset = new Point(offsetX + cosA * x1 + sinA * x2, offsetY - sinA * x1 + cosA * x2);
+                    Point p1 = new Point(-1 * sinA * h + offset.x, -1 * cosA * h + offset.y);
+                    Point p3 = new Point(-1 * cosA * w + offset.x,      sinA * w + offset.y); // original trouble here !
```
lucaro, line 97: what does 'original trouble here !' refer to?

cineast-core/src/main/java/org/vitrivr/cineast/core/util/TextRecognizer.java
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @param coordinate The coordinate of the text
+     * @param frame The frame cotnaining the text
+     * @param quadrilateral If true, the region of the coordinates is regarded as quadrilateral, and quadrilateral cropping is applied
+     * @return The extracted text within the coordinate.
+     */
+    public String recognize(Point[] coordinate, Mat frame, boolean quadrilateral) {
+        Mat cropped = quadrilateral ? quadrilateralCrop(coordinate, frame) : rectangularCrop(coordinate, frame);
+        Imgproc.cvtColor(cropped, cropped, Imgproc.COLOR_RGB2GRAY);
+        Mat blob = Dnn.blobFromImage(cropped, 1/127.5, new Size(100, 32), new Scalar(127.5));
```
lucaro, line 124: What is the meaning behind the 127.5?

lucaro (commented):
Looks good, I added some minor comments regarding details which are unclear to me.

cineast-core/build.gradle
```diff
@@ -150,6 +151,12 @@ dependencies {
     /** JavaCPP. */
     api group: "org.bytedeco", name: "javacpp", version: version_javacpp
     api group: "org.bytedeco.javacpp-presets", name: "ffmpeg", version: version_ffmpeg
+    
+    /** OpenCV. */
+    api group: 'org.openpnp', name: 'opencv', version: '4.5.1-2'
+
+    /** JTS */
+    api group: 'org.locationtech.jts', name: 'jts-core', version: '1.18.1'
```
ppanopticon, line 159: For the sake of consistency, it would make sense to externalize the versions into the gradle.properties file as was done for the remaining dependencies.

ppanopticon (approved):
Looks good to me as far as I could tell. I have two questions:

- Am I understanding this correctly, that you're tracking text boxes over subsequent frames to then determine whether text should be detected or not? That is, we're not just doing detection on the most representative frame.
- Do we have any performance metrics especially for the extraction part?

alexandertheus:
@ppanopticon Yes, as it is currently constituted the scene text is not just extracted from the most representative frame. Namely, detections are done for a subset of the frames, and tracking is applied between adjacent frames at which detections are done.

If I understand you correctly, inference time is a concern, and applying scene text extraction solely on the most representative frame would allow for superior inference time?! If wanted, I can easily extend my code to allow for the option to extract scene text only in the most representative frame.

Performance metrics for the final vitrivr implementation are still to be captured. I will inform you once I know more.

cineast-core/src/main/java/org/vitrivr/cineast/core/util/TextDetector_EAST.java
```diff
@@ -0,0 +1,163 @@
...
+    public Point[][] detect(Mat frame) {
+        Imgproc.cvtColor(frame, frame, Imgproc.COLOR_RGBA2RGB);
+
+        Size size = new Size(frame.width() - (frame.width() % 32), frame.height() - (frame.height() % 32));
+        int H = (int)(size.height / 4);
+        Mat blob = Dnn.blobFromImage(frame, 1.0,size, new Scalar(123.68, 116.78, 103.94), true, false);
+        this.model.setInput(blob);
+        List<Mat> outs = new ArrayList<>(2);
+
+        this.model.forward(outs, this.outNames);
```
alexandertheus, line 121: @silvanheller Yes, the detector and recognizer can be done in a reusable and thread-safe way. However, if I'm not mistaken this would require me to synchronize the code from line 118 to 121. Since this is an especially computationally costly piece of code, this would make the scene text extraction considerably slower.

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractTextRetriever.java
alexandertheus: You are right, I will undo this in the upcoming commit.

cineast-core/src/main/java/org/vitrivr/cineast/core/features/OCRSearch.java
```diff
@@ -1,24 +1,524 @@
...
+   *
+   * threshold_postproc: This is the threshold for the postprocessing stream association step
+   * Strongly urge not to change
+   *
+   * tracker_type: Refers to the tracker which is used
+   * threshold_stream_length: Refers to the amount of consecutive frames a text should minimally appear in
+   * If a text appears in less consecutive frames than the threshold, the text is discarded
    */
+  private static final int rate = 3;
+  private static final int batchSize=16;
```
alexandertheus, line 47: Notable change since the last review: I enabled batch processing for the text detector. It is set to 16 for now, but can easily be changed by altering this parameter.

cineast-core/src/main/java/org/vitrivr/cineast/core/util/TextRecognizer.java
alexandertheus: This is the scaling factor, which every RGB value will be multiplied by in order to scale the input image to a particular range that is needed for this architecture.

cineast-core/src/main/java/org/vitrivr/cineast/core/util/TextDetector_EAST.java
alexandertheus: This comment is removed in the newest version, as it was not meant to stay. For the detector and recognizer implementation, I had initial troubles accessing the right values within the Mat structure, which either gave me out-of-bounds errors, or just erroneous values in a later step of the pipeline.

PR #224, opened 2021-10-18
New Cottontail API, PolyphenyDB

silvanheller:
**Major Changes**

- Updated Cottontail DB connector to version 0.13.x
- Added Polypheny DB connector (experimental)
- Deprecated ADAMpro connector
- Refactored DB integration tests to me more in line with Cineast API
- Changed connection management in Cineast (which is going to be part of a future discussion)

**Breaking**
- Versions of Cottontail DB older than 0.13.0 are no longer supported.


ppanopticon:
@sauterl  @silvanheller Whoever created the DB integration tests should have a look at  _DBIntegrationTest.entriesExistById()_ and _DBBooleanIntegrationTest.entriesExistById()_ I see the following issue:

PersistencyWriter.exists() and PersistencyWriter.idExists() (which are used by those tests) both expect the ID to be a string. However, the entities we create for the test have _long_ and _int_ ID columns. Therefore, these tests will always fail, since Cottontail DB doesn't do type casting. Has this test ever worked?

I see three options: 
- Remove tests or perform them in different setting
- Generalize PersistencyWriter.exists() to use PrimitiveTypeProvider instead of a String
- Change unit tests to use string IDs

Any thoughts?

sauterl:
@silvanheller @ppanopticon My two cents are the generalisation of the `exists()` -- I don't see why this should be hardcoded as a string there.

However, iirc the tests were written by @silvanheller ...

ppanopticon:
> @silvanheller @ppanopticon My two cents are the generalisation of the `exists()` -- I don't see why this should be hardcoded as a string there.
> 
> However, iirc the tests were written by @silvanheller ...

I see your point. But then on the other hand, the data model of Cineast currently expects IDs to be strings, which is hardcoded during entity creation. Adding this level of indirection to the two methods alone would not thus not give us the freedom to choose the type of the ID column.

sauterl:
> I see your point. But then on the other hand, the data model of Cineast currently expects IDs to be strings, which is hardcoded during entity creation. Adding this level of indirection to the two methods alone would not thus not give us the freedom to choose the type of the ID column.

Fair enough, then I'd suggest to go with option C) -- use strings (consistently as we do anywhere else).

silvanheller:
I agree with @sauterl, we can simply make the unit tests expect strings

ppanopticon:
PR is ready for review. 

**Major Changes**

- Updated Cottontail DB connector to version 0.13.x
- Added Polypheny DB connector (experimental)
- Deprecated ADAMpro connector
- Refactored DB integration tests to me more in line with Cineast API
- Changed connection management in Cineast (which is going to be part of a future discussion)

**Breaking**
- Versions of Cottontail DB older than 0.13.0 are no longer supported.

silvanheller:
@sauterl @lucaro @ppanopticon I merged the branch with the current master (which was heavily reformatted in #244), so it should now be ready for review & merging again

cineast-core/src/main/java/org/vitrivr/cineast/core/db/DBSelectorSupplier.java
```diff
@@ -1,7 +0,0 @@
-package org.vitrivr.cineast.core.db;
-
-import java.util.function.Supplier;
-
-public interface DBSelectorSupplier extends Supplier<DBSelector> {
```
lucaro, line 5: This abstraction was introduced deliberately, since there is an `init()` method in both the `Extractor` as well as the `Retriever` which both take a `Supplier` with a different generic type and would therefore clash. Apparently, the `Extractor` now has a mandatory parameter `batchSize` which resolves that clash. This parameter does however appear to be ignored in many cases, so I don't think it should be mandatory if it isn't even consistently respected. Therefore, I suggest to re-introduce this abstraction in order to avoid the clash in generics.

cineast-core/src/main/java/org/vitrivr/cineast/core/db/DBSelector.java
```diff
@@ -26,6 +27,8 @@
 
 public interface DBSelector extends Closeable {
 
+  Logger LOGGER = LogManager.getLogger();
```
lucaro, line 30: Logger is created but never used

cineast-core/src/main/java/org/vitrivr/cineast/core/features/AverageColorRasterReduced11.java
```diff
@@ -14,7 +14,7 @@
 public class AverageColorRasterReduced11 extends AverageColorRaster {
 
   @Override
-  public void init(PersistencyWriterSupplier supply, int batchSize) {
+  public void init(Supplier<PersistencyWriter<?>> supply, int batchSize) {
     /* TODO: Respect batchSize. */
```
lucaro, line 18: here, for example...

lucaro (changes requested):

silvanheller:
Just FYI @ppanopticon none of the unit tests currently work for polypheny

ppanopticon:
> Just FYI @ppanopticon none of the unit tests currently work for polypheny

I have fixed a few issues in the DB connector and adjusted the unit tests. Main reasons for failure:

- Polypheny DB requires a PK field, which was not specified in the Unit Test (i.e., all tests failed)
- Minor issues in the implementation

Currently, two tests are failing:
- _testBetweenANDInQuery():_  Here the issue seems to lie in the abstract implementation because I have actually not implemented this method.
- _entriesExistById():_ There seems to be a Polypheny DB issue here. I have to check with Marco.

cineast-core/src/main/java/org/vitrivr/cineast/core/db/DBSelectorSupplier.java
ppanopticon: I resolved this and removed `batchSize` from the `init()` method. It is now a property of the `PersistencyWriter` itself.

cineast-core/src/main/java/org/vitrivr/cineast/core/db/DBSelector.java
ppanopticon: I beg to disagree: `CottontailSelector`, `PolyphenySelector` and `DataSource` all use this LOGGER.

ppanopticon:
@lucaro @sauterl @silvanheller  (no idea who built this) I noticed that there is a version conflict between the ProtoBuf / gRPC version used by Cottontail DB and the one used by Cineast itself, which manifests itself by a crashing gRPC service in Cineast upon startup.

Would it be possible and safe to bump the versions for Cineast?

lucaro:
I don't see why not. I built the gRPC API for cineast a while ago, mostly as an attempt to cleanly specify all the calls and messages from scratch and also to make it easier to interface with other systems. It might already be outdated again though and nobody has used it since.

ppanopticon:
I fixed the `exists()` unit test by changing the query, since I do not have any influence over how and when the causing bug will be fixed in Polypheny DB.

As to handling of connections between Cottontail DB and Polypheny DB: I don't see an inconsistency here. JDBC connections are closed right away after an instance of a DAO is closed(). This is a bit different to gRPC, because the general recommendation for `ManagedChannels` is to keep them open for multiple calls. This is why there is some kind of connection pooling in the `CottontailWrapper`.

I don't have any troubles running either unit test locally.

cineast-core/src/test/java/org/vitrivr/cineast/core/db/dao/MetadataTest.java
```diff
@@ -96,9 +96,6 @@ void setupTest() {
   @AfterEach
   void tearDownTest() {
     dropTables();
-
-    /* Close all readers / writers. */
-    CineastIOUtils.closeQuietly(this.objWriter, this.segWriter, this.objReader, this.segReader);
```
silvanheller: This should probably be moved to the @AfterAll method

ppanopticon: Did that.

ppanopticon:
I have addressed all the issues. This PR is ready for final review.

I will open an issue regarding DB connection handling in Cineast.

cineast.json
```diff
@@ -1,9 +1,9 @@
 {
   "database": {
     "host": "localhost",
-    "selector": "COTTONTAIL",
-    "writer": "COTTONTAIL",
-    "port": 1865,
+    "selector": "POLYPHENY",
+    "writer": "POLYPHENY",
+    "port": 80,
```
lucaro, line 6: I don't think we want to make this the default, so this should be changed back to cottontail before being merged.

lucaro (changes requested):
Looks all good, with the minor exception of the changes to the default config.

lucaro (dismissed):

silvanheller:
LGTM, thanks to everyone for all the work! I've tested this on the VBS data manually a bit and found no issues

PR #231, opened 2021-11-08
Fix and extend REST API functionality for features

simonpeterhans:
Adds two new REST API endpoints to obtain features and fixes existing endpoints. 

Contents:
- Update some paths that were overseen in `APIEndpoint` to comply with Javalin 4
 - Fix `retrieveFeaturesForAllCategories()` to properly return features from all categories instead of returning after the first one of each category
- Add endpoint to obtain all features in an entity/table (either for all indexed objects or a list of IDs)
- Add endpoint to obtain all features in entities of a category (either for all indexed objects or a list of IDs)


sauterl (approved):

docs/openapi.json
```diff
@@ -1,1632 +1,1742 @@
...
-        } ],
-        "responses" : {
-          "200" : {
-            "description" : "OK",
-            "content" : {
-              "application/json" : {
-                "schema" : {
-                  "$ref" : "#/components/schemas/MediaObjectMetadataQueryResult"
-                },
-                "exampleSetFlag" : false
```
sauterl, line 68: This should have been removed

sauterl (changes requested):
The OpenAPI specs seem to be off. They should be regenerated

sauterl (approved):

PR #232, opened 2021-11-15
HOTFIX: removing jcenter from repositories

silvanheller:
JCenter is currently down, which has led to me discovering that we should not be using it anymore anyway: https://blog.gradle.org/jcenter-shutdown


lucaro:
If all dependencies are also available from other repositories, it's good to be rid of unnecessary ones. 👍

PR #238, opened 2021-12-09
Visual-text co-embedding video extraction

Spiess:
Implemented the extraction of the visual-text co-embedding feature for videos, such that not only the most representative frame for each shot is considered, but all frames mean-pooled in the encoding space.

Additional minor cleanup refactoring in the MotionHistogramCalculator.

### DIFF
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/features/VisualTextCoEmbedding.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/features/VisualTextCoEmbedding.java
@@ -6,6 +6,7 @@ import java.awt.image.BufferedImage;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import org.tensorflow.SavedModelBundle;
 import org.tensorflow.Tensor;
 import org.tensorflow.ndarray.NdArrays;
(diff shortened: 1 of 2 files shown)


cineast-core/src/main/java/org/vitrivr/cineast/core/features/VisualTextCoEmbedding.java
```diff
@@ -87,15 +89,27 @@ public VisualTextCoEmbedding() {
...
+      // Segment is video
+      List<BufferedImage> frames = shot.getVideoFrames().stream()
+          .map(frame -> frame.getImage().getBufferedImage())
+          .collect(Collectors.toList());
 
-    if (image != null) {
-      float[] embeddingArray = embedImage(image);
+      float[] embeddingArray = embedVideo(frames);
       this.persist(shot.getId(), new FloatVectorImpl(embeddingArray));
+    } else if (shot.getMostRepresentativeFrame() != VideoFrame.EMPTY_VIDEO_FRAME) {
```
silvanheller, line 105: in general, early exit is preferred over else if

silvanheller (changes requested):

silvanheller (approved):

cineast-core/src/main/java/org/vitrivr/cineast/core/features/VisualTextCoEmbedding.java
```diff
@@ -203,6 +223,77 @@ private void initializeVisualEmbedding() {
     }
   }
 
+  private float[] embedVideo(List<BufferedImage> frames) {
```
lucaro, line 226: I don't think it's ideal to use the BufferedImage here, since that circumvents the frame caching mechanism. You could just as well use the MultiImage and access the buffered one once you need it, without having to load all of them first.

lucaro (changes requested):

lucaro (approved):

PR #243, opened 2022-01-03
InceptionResnetV2 feature for Query-by-Example

Spiess:
Refactored InceptionResnetV2 feature encoding into its own separate feature from inside VisualTextCoEmbedding.
This allows it to be used as a feature for query-by-example that is less abstracted towards semantic content than the visual-text co-embedding.

### DIFF
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
@@ -1,0 +1,205 @@
+package org.vitrivr.cineast.core.features;
+
+import java.awt.geom.AffineTransform;
+import java.awt.image.AffineTransformOp;
+import java.awt.image.BufferedImage;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.tensorflow.SavedModelBundle;
+import org.tensorflow.Tensor;
+import org.tensorflow.ndarray.Shape;
+import org.tensorflow.ndarray.buffer.DataBuffers;
+import org.tensorflow.ndarray.buffer.FloatDataBuffer;
+import org.tensorflow.types.TFloat32;
+import org.vitrivr.cineast.core.config.QueryConfig;
+import org.vitrivr.cineast.core.config.ReadableQueryConfig;
+import org.vitrivr.cineast.core.config.ReadableQueryConfig.Distance;
+import org.vitrivr.cineast.core.data.FloatVectorImpl;
+import org.vitrivr.cineast.core.data.frames.VideoFrame;
+import org.vitrivr.cineast.core.data.score.ScoreElement;
+import org.vitrivr.cineast.core.data.segments.SegmentContainer;
+import org.vitrivr.cineast.core.features.abstracts.AbstractFeatureModule;
+
+public class InceptionResnetV2 extends AbstractFeatureModule {
+
+  public static final int ENCODING_SIZE = 1536;
+  private static final String TABLE_NAME = "features_inceptionresnetv2";
+  private static final Distance DISTANCE = ReadableQueryConfig.Distance.euclidean;
+
+  private static final Logger LOGGER = LogManager.getLogger();
+
+  /**
+   * Required dimensions of visual embedding model.
+   */
+  public static final int IMAGE_WIDTH = 299;
+  public static final int IMAGE_HEIGHT = 299;
+
+  /**
+   * Resource paths.
+   */
+  private static final String MODEL_PATH = "resources/VisualTextCoEmbedding/inception_resnet_v2_weights_tf_dim_ordering_tf_kernels_notop";
+
+  /**
+   * Model input and output names.
+   */
+  public static final String INPUT = "input_1";
+  public static final String OUTPUT = "global_average_pooling2d";
+
+  /**
+   * InceptionResNetV2 pretrained on ImageNet: https://storage.googleapis.com/tensorflow/keras-applications/inception_resnet_v2/inception_resnet_v2_weights_tf_dim_ordering_tf_kernels_notop.h5
+   */
+  private static SavedModelBundle model;
+
+  public InceptionResnetV2() {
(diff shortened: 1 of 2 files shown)


silvanheller (approved):
Looks good to me

cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+  }
+
+  // TODO: Move image util related functions to a dedicated util class
+
+  /**
+   * Rescales a buffered image using bilinear interpolation.
+   */
+  public static BufferedImage rescale(BufferedImage image, int width, int height) {
+    BufferedImage scaledImage = new BufferedImage(width, height, image.getType());
```
lucaro, line 157: We usually use another library for image operations, since `awt` might (in some rare instances) not be available on headless machines. Currently, the `Thumbnailator` is used (although I might want to replace that at some point since there are other libraries that are more actively maintained).

lucaro (changes requested):
I generally agree that it makes much more sense to have the output of the resnet as a feature directly. I only have one minor comment regarding the method used for image scaling, see inline comment.

cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
Spiess: I'm not sure I understand your concern correctly; I had a quick look at Thumbnailator and it is also [depending on awt](https://github.com/coobird/thumbnailator/blob/b97f8fba767a5a0b5378bdd9cb3c81f39b295b00/src/main/java/net/coobird/thumbnailator/Thumbnailator.java#L27).

lucaro: One more reason to replace it then, although the dependency might only be for compatibility to the data structures, I'd have to check. The main point is: there is a more readable and potentially more efficient way to achieve the same result which is (ideally) consistently used by all the other features, so I don't see a reason not to use it here as well.

Spiess: Unfortunately, I have discovered that the `Thumbnailator` can only resize images while preserving the original aspect ratio, something that is not compatible with the input requirements of InceptionResnetV2. Since the replacement of the `Thumbnailator` is an entirely different issue to this pull request, I suggest merging this as-is and opening a new issue for the replacement and unification of image utilities.

lucaro: Nope, if you want to change the aspect ratio, simply use `forceSize(w, h)` or `size(w, h).keepAspectRatio(false)` as stated in the documentation. I also just saw that the maintenance appears to have picked up, so there's probably no need to change it after all. An update might be in order though.

silvanheller (dismissed):
I bumped the version number to v3.6.2, since that's what the next minor release would be

cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
Spiess: My bad, I had misunderstood the documentation.

silvanheller (approved):

cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
```diff
@@ -0,0 +1,202 @@
...
+  public void processSegment(SegmentContainer shot) {
+    // Return if already processed
+    if (phandler.idExists(shot.getId())) {
+      return;
+    }
+
+    // Case: segment contains image
+    if (shot.getMostRepresentativeFrame() != VideoFrame.EMPTY_VIDEO_FRAME) {
+      BufferedImage image = shot.getMostRepresentativeFrame().getImage().getBufferedImage();
+
```
lucaro, line 71: It stands to reason if the feature should aggregate the frame vectors rather than using the vector of the aggregated frame. It does however certainly work like this.

lucaro (approved):

cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
Spiess: If you are referring to the case where the provided shot contains a video segment rather than just a single image, I agree that aggregating the resulting feature vectors rather than using the most representative frame would probably lead to better results, however, if I understand our current use case for this correctly, it is more targeted towards image-content-based similarity detection / more-like-this queries for images.

Of course, this functionality can be easily implemented if the need arises.

lucaro: If you use all frames of a segment and aggregate over them, the behavior in the case of a single image will be equivalent to the current implementation but consider more information in the case of video. In most cases, it will however probably not make a huge difference.

silvanheller (dismissed):

cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
```diff
@@ -0,0 +1,255 @@
...
+  @Override
+  public void processSegment(SegmentContainer sc) {
+    // Return if already processed
+    if (phandler.idExists(sc.getId())) {
+      return;
+    }
+
+    // Case: segment contains video frames
+    if (!(sc.getVideoFrames().size() > 0 && sc.getVideoFrames().get(0) == VideoFrame.EMPTY_VIDEO_FRAME)) {
```
lucaro, line 71: In general, never do `collection.size() > 0`, always do `!collection.isEmpty()`.

lucaro (dismissed):

cineast-core/src/main/java/org/vitrivr/cineast/core/features/InceptionResnetV2.java
Spiess: You are absolutely right, this is a mistake I made while figuring out how to check for video content in a segment container. I will fix it right away.

silvanheller (approved):

lucaro (approved):

PR #244, opened 2022-01-04
v3.6.1: Java 11, Embedding Improvements, Complete Reformatting

silvanheller:
# Changes
* Updates to text co-embedding by @Spiess 
* Complete reformatting of codebase by @Spiess to adhere to codebase guidelines

I'd like to merge this PR before #224, because #224 targets master and I think it makes more sense to first merge the reformatting. I'll take care of making sure #224 is merged with this branch, since that might be quite some work due to the reformatting.


lucaro (approved):
Going by the commit log, this should not contain any semantic changes but only code maintenance. I'd therefore be in favor to merge this as soon as feasible in order to not mess up subsequent merges.

Spiess (approved):

PR #245, opened 2022-01-05
InceptionResnetV2 feature for Query-by-Example

Spiess:
Refactored InceptionResnetV2 feature encoding into its own separate feature from inside VisualTextCoEmbedding.
This allows it to be used as a feature for query-by-example that is less abstracted towards semantic content than the visual-text co-embedding.

Reopen of #243 due to merge error.


silvanheller (approved):
Approving since we discussed everything in #243

PR #252, opened 2022-02-04
Fix remaining Java build warnings

Spiess:
Fixes the remaining java build warnings (fixable from within Cineast).


Spiess:
I would suggest not bumping the version number again, since we did not issue a new release since the previous pull request was merged, and instead issuing a new release after.

Issue #208, opened 2021-06-08
Switch publishing system


silvanheller:
We should use a different plugin to publish our jars so we can upgrade to Gradle 7. See [Gradle Docs](https://docs.gradle.org/6.7.1/userguide/upgrading_version_5.html#legacy_publication_system_is_deprecated_and_replaced_with_the_publish_plugins). Additionally, it seems that v3.2.0 has broken the jar publishing process (see CI status)

Spiess:
Judging by the CI status, this was fixed by #253.

PR #253, opened 2022-02-04
Removed maven plugin

Spiess:
Removed maven plugin, possibly fixing #208 and added @Spiess as developer to publish task.


silvanheller:
Well I think we probably want to have an alternative publishing system, no?

Spiess:
There is already a replacement publishing system in place with [`maven-publish`](https://github.com/vitrivr/cineast/blob/c0c4de72516dcd8ab306c685afe89ba82c12fb81/build.gradle#L17).
As far as I can tell, the `maven` plugin no longer exists and caused the publishing process to fail, but I will need to find a way to test this locally to confirm.

silvanheller:
So `gradle publish` should simply use `maven-publish` automatically?

Spiess:
That appears to be the case for the [Cottontail DB Proto](https://github.com/vitrivr/cottontaildb-proto) repo at least, but maybe @ppanopticon has further insight.

Spiess:
I have tested and confirmed the fix through publish to local maven cache.

ppanopticon (approved):
Looks good to me!

Issue #258, opened 2022-02-10
Large amounts of Cottontail connection and closing messages


Spiess:
Starting Cineast with or without connected Cottontail DB causes large amounts of Cottontail connection messages, either caused by a logging bug or several connections actually being established.
Stopping Cineast causes equally many Connection closing messages to be logged.
```
No config path given, loading default config cineast.json
2022-02-10 09:31:14.322 [main] INFO o.v.c.s.c.Config - Config file loaded!
2022-02-10 09:31:14.337 [main] DEBUG o.v.c.c.d.c.CottontailWrapper - Starting to connect to Cottontail DB at localhost:1865
2022-02-10 09:31:14.451 [main] INFO o.v.c.c.d.c.CottontailWrapper - Connected to Cottontail DB in 115 ms at localhost:1865
2022-02-10 09:31:14.583 [main] WARN o.v.c.c.d.c.CottontailWrapper - Could not connect to Cottontail at localhost:1865
2022-02-10 09:31:14.591 [main] DEBUG o.v.c.c.d.c.CottontailWrapper - Starting to connect to Cottontail DB at localhost:1865
2022-02-10 09:31:14.592 [main] INFO o.v.c.c.d.c.CottontailWrapper - Connected to Cottontail DB in 0 ms at localhost:1865
2022-02-10 09:31:14.598 [main] WARN o.v.c.c.d.c.CottontailWrapper - Could not connect to Cottontail at localhost:1865
2022-02-10 09:31:14.620 [main] ERROR o.v.c.c.d.DBSelector - Failed to close TupleIterator!
io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:262) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:243) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:156) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.grpc.DDLGrpc$DDLBlockingStub.entityDetails(DDLGrpc.java:856) ~[cineast-api-3.7.0-all.jar:1.xx]
[...]

Issue #257, opened 2022-02-10
Cineast CLI fails to start


Spiess:
On the current master branch, Cineast never reaches the CLI and does not accept any CLI commands.
Possibly caused by #254.

lucaro:
I cannot reproduce this, could you elaborate?

Spiess:
On a clean repository clone, running the shadowJar build cineast-api Jar file no CLI commands can be issued.
Tested on macOS and Linux.

Abbreviated output example (`test` and `help` were entered):
```
...
2022-02-10 11:02:27.729 [main] INFO o.v.c.a.GRPCEndpoint - Starting GRPC Endpoint at port 4570
Exception in thread "main" java.lang.AbstractMethodError: Receiver class io.grpc.netty.shaded.io.grpc.netty.NettyServerBuilder$NettyClientTransportServersBuilder does not define or inherit an implementation of the resolved method 'abstract io.grpc.internal.InternalServer buildClientTransportServers(java.util.List)' of interface io.grpc.internal.ServerImplBuilder$ClientTransportServersBuilder.
	at io.grpc.internal.ServerImplBuilder.build(ServerImplBuilder.java:242)
	at io.grpc.internal.AbstractServerImplBuilder.build(AbstractServerImplBuilder.java:173)
	at org.vitrivr.cineast.api.GRPCEndpoint.start(GRPCEndpoint.java:32)
	at org.vitrivr.cineast.api.Main.main(Main.java:32)
test
help
^C2022-02-10 11:02:32.814 [Thread-23] INFO o.v.c.c.d.c.CottontailWrapper - Closing connection to Cottontail DB.
...
```

ppanopticon:
Indeed if unhandled exceptions stemming from one of the following lines bubble up, then the CLI won't start. Wrapping those in a try...catch would be a solution. However, chances are that if initialisation fails, the CLI will not be able to do useful work either.

```java
      /* Start API endpoint. */
      APIEndpoint.getInstance().start();

      /* Start gRPC endpoint. */
      GRPCEndpoint.start();

      /* Initialize Monitoring */
      PrometheusServer.initialize();
```

Issue #256, opened 2022-02-10
Error with large stack trace on startup if no CottontailDB connected


Spiess:
Starting Cineast without a CottontailDB connected causes an error with a large stack trace:
```
[main] ERROR o.v.c.c.d.DBSelector - Failed to close TupleIterator!
io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:262) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:243) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:156) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.grpc.DDLGrpc$DDLBlockingStub.entityDetails(DDLGrpc.java:856) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.client.SimpleClient.about$lambda-38$lambda-37(SimpleClient.kt:430) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.Context.call(Context.java:551) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.client.SimpleClient.about$lambda-38(SimpleClient.kt:428) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.Context.call(Context.java:551) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.client.SimpleClient.about(SimpleClient.kt:426) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.client.SimpleClient.about(SimpleClient.kt:440) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cineast.core.db.cottontaildb.CottontailSelector.existsEntity(CottontailSelector.java:345) [cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cineast.core.db.dao.reader.TagReader.initCache(TagReader.java:162) [cineast-api-3.7.0-all.jar:1.xx]
[...]

Issue #255, opened 2022-02-10
Error in CottontailDB connection if schema does not already exist


Spiess:
Starting Cineast connected to a CottontailDB instance without the schema warren.cineast already in the database causes an error with a large stack trace:
```
[main] ERROR o.v.c.c.d.DBSelector - Failed to close TupleIterator!
io.grpc.StatusRuntimeException: NOT_FOUND: [2, 53715eb5-2843-4c78-9ec0-c3058498c443] Execution of null failed: Schema 'warren.cineast' does not exist!
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:262) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:243) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:156) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.grpc.DDLGrpc$DDLBlockingStub.entityDetails(DDLGrpc.java:856) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.client.SimpleClient.about$lambda-38$lambda-37(SimpleClient.kt:430) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.Context.call(Context.java:551) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.client.SimpleClient.about$lambda-38(SimpleClient.kt:428) ~[cineast-api-3.7.0-all.jar:1.xx]
	at io.grpc.Context.call(Context.java:551) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.client.SimpleClient.about(SimpleClient.kt:426) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cottontail.client.SimpleClient.about(SimpleClient.kt:440) ~[cineast-api-3.7.0-all.jar:1.xx]
	at org.vitrivr.cineast.core.db.cottontaildb.CottontailSelector.existsEntity(CottontailSelector.java:345) [cineast-api-3.7.0-all.jar:1.xx]
[...]

PR #259, opened 2022-02-10
Fixes for Cottontail DB Adapter

ppanopticon:
Addresses #255, #256, #257 #258, #257 as requested by @Spiess

### DIFF
--- a/cineast-api/src/main/java/org/vitrivr/cineast/api/GRPCEndpoint.java
+++ b/cineast-api/src/main/java/org/vitrivr/cineast/api/GRPCEndpoint.java
@@ -18,7 +18,7 @@ public class GRPCEndpoint {
 
   private static final Logger LOGGER = LogManager.getLogger();
 
-  public static void start() {
+  public static void start() throws IOException {
 
     if (!Config.sharedConfig().getApi().getEnableGRPC()) {
       return;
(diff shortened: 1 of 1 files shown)


ppanopticon:
These fixes should address the issues reported by @Spiess 

I'm not if removing ADAMpro should als be included in this PR.

Spiess:
Thank you for the quick fixes. Once you have confirmed that the many database channels are indeed intended on #258 I will approve the mere.

cineast-api/src/main/java/org/vitrivr/cineast/api/Main.java
```diff
@@ -27,12 +27,26 @@ public static void main(String[] args) {
...
+    /* Start API endpoint. */
+    try {
+      APIEndpoint.getInstance().start();
+    } catch (Throwable e) {
+      System.err.println("Failed to initialize API endpoint due to an exception: " + e.getMessage());
+    }
+
+    /* Start gRPC endpoint. */
+    try {
+      GRPCEndpoint.start();
```
lucaro, line 39: If you do error checking here, the block in https://github.com/vitrivr/cineast/blob/cecc4ac9c69f4806a1ab04d9c9888d432fd6eb49/cineast-api/src/main/java/org/vitrivr/cineast/api/GRPCEndpoint.java#L36 can be removed (and should be, as failing silently is never a good strategy).

lucaro (changes requested):

cineast-api/src/main/java/org/vitrivr/cineast/api/Main.java
ppanopticon: Done!

ppanopticon:
> Thank you for the quick fixes. Once you have confirmed that the many database channels are indeed intended on #258 I will approve the merge.

This is intended. The CottontailWrapper class is lightweight and re-created many times. The only thing I'm not sure about is whether pinging the endpoint is necessary every, single time.

lucaro (approved):

PR #261, opened 2022-02-11
Fixed two issues related to feature extraction

ppanopticon:
**Changes:** 
- Initialisation of Writers now takes place in construction
- DatabaseConfig.getSelectorSupplier() now respects selector config instead of writer config.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ allprojects {
     group = 'org.vitrivr'
 
     /* Our current version, on dev branch this should always be release+1-SNAPSHOT */
-    version = '3.7.0'
+    version = '3.7.3'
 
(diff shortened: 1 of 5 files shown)


lucaro:
Is the `init()` method still used for any of the writers? If not, we might as well get rid of it.

Spiess (approved):
Works great, thanks for fixing so quickly.

ppanopticon:
> Is the `init()` method still used for any of the writers? If not, we might as well get rid of it.

I agree! But I believe that this should be part of the refactoring proposed in #251. Since this is a larger undertaking, and the issues at hand are currently blockers for @Spiess, I'd suggest to move forward without that change at this point.

lucaro (approved):

PR #268, opened 2022-02-25
Null-safety for temporal queries in WS API

silvanheller:
This PR makes sure that null is actually an acceptable parameter for the websocket API


Spiess (approved):
LGTM, thanks

PR #274, opened 2022-03-07
Fixing NearestNeighborRows only returning id column

silvanheller:
The `DBSelector` Interface had two abstractions for nearest-neighbor search, one of which only returns ids & distances, the other one returns the full row. The second one is used e.g by the `AverageColorRaster` Feature which has the schema `id`, `hist`, `raster` and requires the raster for score-computation. 

This PR fixes the cottontail implementation to avoid NPEs in `AverageColorRaster`. Thanks to @benzvera for pointing out the bug.
It also optimizes the `countDistinctValues` command, which only needs to retrieve the column for which it counts distinct values.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ allprojects {
     group = 'org.vitrivr'
 
     /* Our current version, on dev branch this should always be release+1-SNAPSHOT */
-    version = '3.8.3'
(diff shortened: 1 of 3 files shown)


ppanopticon (approved):
Looks good.

PR #275, opened 2022-03-11
Adding ProvidedOCR feature, option to disable transactions during import

silvanheller:
Adding a ProvidedOCR feature which is explicit about being for imported data, adding an option to disable transactions during import

### DIFF
--- a/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/TextRetrievalCommand.java
+++ b/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/TextRetrievalCommand.java
@@ -35,7 +35,6 @@ public class TextRetrievalCommand implements Runnable {
     retrievers.add(new OCRSearch());
     retrievers.add(new AudioTranscriptionSearch());
     retrievers.add(new DescriptionTextSearch());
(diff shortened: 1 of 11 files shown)


cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/TextRetrievalCommand.java
```diff
@@ -35,7 +35,6 @@ public void run() {
     retrievers.add(new OCRSearch());
     retrievers.add(new AudioTranscriptionSearch());
     retrievers.add(new DescriptionTextSearch());
-    retrievers.add(new TagsFtSearch());
```
lucaro, line 38: doesn't this one exist anymore?

silvanheller: It's not part of the standard functionality a user can reasonably expect from any Cineast Instance since it searches a separate entity with a full-text index that has to be externally created.

lucaro: fair enough.

lucaro (approved):

PR #276, opened 2022-03-11
Removed deprecated ADAMpro database connector.

Spiess:
Resolves #270.

Issue #270, opened 2022-03-03
Remove all deprecated ADAMpro Code


silvanheller:
All references & classes related to ADAMpro can be removed

PR #278, opened 2022-03-13
Cleanup: Removing ADAMpro dependencies, unused Code

silvanheller:
This PR removes the adampro grpc dependency as a followup to #270, #276 , which requires changes in a lot of places, since some dependencies were only available thanks to that dependency apparently. Additionally, it removes a lot of unused code such as:
- Code from a BSc Thesis from 2016 which still had hardcoded paths (?)
- Import-Code from 2019/2020/2021 which is not needed anymore and should anyway be in a separate repository. The current import-code now supports only LIRE and the Cineast-generated JSON Features.
- Evaluation-Code from 2017 which was not really compatible with the architecture anymore.
- Removes Motion-Features as they were never used and not benchmarked (also see #277)


cineast-api/src/main/java/org/vitrivr/cineast/api/websocket/handlers/queries/TemporalQueryMessageHandler.java
```diff
@@ -10,6 +9,8 @@
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
+
+import it.unimi.dsi.fastutil.ints.Int2ObjectLinkedOpenHashMap;
```
lucaro, line 13: FastUtil was imported via ADAMpro and was used in some places, despite Cineast traditionally using Trove for such purposes. They both do very similar things. I don't know which one is better, I've mostly worked with Trove in the past. In the context of such a cleanup effort, I think it would be worth to just use one of the two.

cineast-core/src/main/java/org/vitrivr/cineast/core/util/ocr/ThreadLocalObjectCache.java
```diff
@@ -1,4 +1,4 @@
-package org.vitrivr.cineast.core.util;
+package org.vitrivr.cineast.core.util.ocr;
```
lucaro, line 1: This one is not specifically something for OCR, even though it is currently only used there.

cineast-core/src/main/java/org/vitrivr/cineast/core/util/ocr/HungarianAlgorithm.java
```diff
@@ -1,4 +1,4 @@
-package org.vitrivr.cineast.core.util;
+package org.vitrivr.cineast.core.util.ocr;
```
lucaro, line 1: This is not OCR-specific functionality, it's also used for the skeleton stuff for example.

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/RetrieveCommand.java
```diff
@@ -21,17 +20,11 @@
...
 
-  @Option(name = {"-e", "--export"}, title = "Export", description = "Indicates whether the results should be exported. Defaults to false.")
-  private boolean export = false;
-
   @Option(name = {"-r", "--relevantSegments"}, title = "Relevant segments", description = "Comma separated list of segment IDs to which the query is to be limited.")
   private String relevantSegments;
 
   public void run() {
     final ContinuousRetrievalLogic retrieval = new ContinuousRetrievalLogic(Config.sharedConfig().getDatabase());
-    if (export) {
```
lucaro, line 32: Having this might be useful for testing and evaluation purposes, don't you think?

lucaro (changes requested):

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/RetrieveCommand.java
silvanheller: Not really, i think for testing / evaluation, you'd want to use the APIs. For internal purposes, this was not really using a sensible abstraction (as evidenced by the fact that all evaluations of cineast / cottontail in past years have not used this functionality)

cineast-api/src/main/java/org/vitrivr/cineast/api/websocket/handlers/queries/TemporalQueryMessageHandler.java
silvanheller: I've removed fastutil again and replaced the two usages with their trove-equivalent

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/cli/RetrieveCommand.java
lucaro: ok, fair enough.

lucaro (approved):

lucaro (dismissed):

cineast-core/src/main/java/org/vitrivr/cineast/core/db/dao/reader/AbstractMetadataReader.java
```diff
@@ -105,11 +116,17 @@ public AbstractMetadataReader(DBSelector selector, String tableName, String idCo
   }
 
   public List<R> findBySpec(MetadataAccessSpecification... spec) {
```
Spiess, line 118: These functions appear to only be used in tests. Would it be possible to add a query ID parameter and set it to null (or something more appropriate) in the test instead?

cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailSelector.java
```diff
@@ -125,32 +130,36 @@ public void close() { /* No op. */ }
...
   public List<Map<String, PrimitiveTypeProvider>> getRows(String fieldName, Iterable<PrimitiveTypeProvider> values) {
     final Object[] mapped = StreamSupport.stream(values.spliterator(), false).map(PrimitiveTypeProvider::toObject).toArray();
-    final Query query = new Query(this.fqn).select("*", null).where(new Literal(fieldName, "IN", mapped));
-    try {
-      return processResults(this.cottontail.client.query(query));
-    } catch (StatusRuntimeException e) {
-      LOGGER.warn("Error occurred during query execution in getRows(): {}", e.getMessage());
-      return new ArrayList<>(0);
-    }
+    return getRowsHelper(fieldName, "IN", mapped, "get-rows-in-iterable");
```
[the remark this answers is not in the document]
silvanheller: good point - but for example in the Rest API, there's not always a query ID so I held off on forcing all `DBSelector` methods to provide one.

cineast-core/src/main/java/org/vitrivr/cineast/core/db/dao/reader/AbstractMetadataReader.java
silvanheller: I guess it's a preference thing - like this, it's clear that the queryID is optional without looking at the implementation or usages of the code

PR #281, opened 2022-03-20
improving robustness when feature entity is missing

silvanheller:
In rare cases, the `segmenttags` entity is not available when an API request is made. In this case, Cineast after this PR simply returns an empty String instead of failing the entire request.

### DIFF
--- a/cineast-api/src/main/java/org/vitrivr/cineast/api/util/QueryUtil.java
+++ b/cineast-api/src/main/java/org/vitrivr/cineast/api/util/QueryUtil.java
@@ -267,7 +267,7 @@ public class QueryUtil {
       return _return;
     } catch (Exception e) {
       LOGGER.error("Exception while looking up tags", e);
-      return new ArrayList<>();
+      return _return;
     }
   }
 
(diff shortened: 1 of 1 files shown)


cineast-api/src/main/java/org/vitrivr/cineast/api/util/QueryUtil.java
```diff
@@ -261,10 +261,14 @@ private static void retrieveAndWeight(ContinuousRetrievalLogic continuousRetriev
...
-
-    rows.forEach(row -> _return.add(row.get(SegmentTags.TAG_ID_QUALIFIER).getString()));
-    return _return;
+    try {
+      List<Map<String, PrimitiveTypeProvider>> rows = selector.getRows(GENERIC_ID_COLUMN_QUALIFIER, ids);
+      rows.forEach(row -> _return.add(row.get(SegmentTags.TAG_ID_QUALIFIER).getString()));
+      return _return;
+    } catch (Exception e) {
+      LOGGER.error("Exception while looking up tags", e);
+      return new ArrayList<>();
```
lucaro, line 270: Since you are already allocating `_return`, you can return it after the try/catch block in either case

lucaro (changes requested):

lucaro (approved):

PR #282, opened 2022-03-20
creating index on id-column by default

silvanheller:
This PR adds an index on the `id` column for all entities by default. This is desirable for mlt requests, and for users which seek more information about the features of a given segment.

### DIFF
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
@@ -5,10 +5,13 @@ import static org.vitrivr.cineast.core.db.setup.AttributeDefinition.AttributeTyp
 import static org.vitrivr.cineast.core.db.setup.AttributeDefinition.AttributeType.TEXT;
 import static org.vitrivr.cineast.core.db.setup.AttributeDefinition.AttributeType.VECTOR;
 import static org.vitrivr.cineast.core.util.CineastConstants.GENERIC_ID_COLUMN_QUALIFIER;
+import static org.vitrivr.cottontail.grpc.CottontailGrpc.IndexType.HASH;
+import static org.vitrivr.cottontail.grpc.CottontailGrpc.IndexType.HASH_UQ;
 
 import io.grpc.StatusRuntimeException;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
```diff
@@ -205,7 +208,9 @@ public boolean createFeatureEntity(String featureEntityName, boolean unique, int
     final AttributeDefinition[] attributes = Arrays.stream(featureNames)
         .map(s -> new AttributeDefinition(s, VECTOR, length))
         .toArray(AttributeDefinition[]::new);
-    return this.createFeatureEntity(featureEntityName, unique, attributes);
+    var feature = this.createFeatureEntity(featureEntityName, unique, attributes);
+    this.createIndex(featureEntityName, GENERIC_ID_COLUMN_QUALIFIER, unique ? IndexType.HASH_UQ : IndexType.HASH, null);
```
lucaro, line 212: It might be worth having indexes for the other fields as well, at least for the object id.

lucaro (changes requested):

cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
silvanheller: Feature entities as create in this method only have the segment id. the object id is indexed in the segments entity https://github.com/vitrivr/cineast/blob/a2ced2934afdc947d56f471f59bf220dce0c09ef/cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java#L128

If you mean that we should index the field partially, i'm not sure if Cottontail supports that.

lucaro: I was specifically referring to the segments entity, the comment might have ended up on the wrong line. It might be worth indexing all the columns of the 'book-keeping' entities since they might be used for different lookups.

silvanheller: Yes, that's already being done, see https://github.com/vitrivr/cineast/blob/a2ced2934afdc947d56f471f59bf220dce0c09ef/cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java#L145

lucaro (approved):

PR #285, opened 2022-03-28
Feature/pose queries

lucaro:
This adds the relevant functionality for pose-based queries.


cineast-api/src/main/java/org/vitrivr/cineast/api/messages/query/TemporalQuery.java
```diff
@@ -76,4 +77,8 @@ public TemporalQueryConfig getTemporalQueryConfig() {
   public MessageType getMessageType() {
     return MessageType.Q_TEMPORAL;
   }
+
+  public void setTimeDistances(List<Float> distances) {
```
silvanheller, line 81: If you have merged vitrivr-ng correctly, this field is not required. It is present in the `TemporalQueryConfig` (it's possible that this is my bad). If you open a PR for the vitrivr-ng branch, I can check that.

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/resolvers/FileSystemThumbnailResolver.java
```diff
@@ -35,7 +35,7 @@ public ResolutionResult resolve(String segmentId) {
     File[] candidates = dir.listFiles(new FilenameFilter() {
       @Override
       public boolean accept(File dir, String name) {
-        return name.startsWith(segmentId);
+        return name.startsWith("shot" + split[1] + "_" + split[2]);
```
silvanheller, line 38: While I haven't used these resolvers in the past, I'm not sure why these changes are required. Could you add a small comment, so we can check if the documentation or the default config requires an update?

cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
```diff
@@ -143,8 +141,8 @@ public boolean createSegmentEntity() {
       this.cottontail.client.create(entity);
 
       /* Create indexes. */
-      this.createIndexWithoutSchemaName(MediaSegmentDescriptor.ENTITY, MediaSegmentDescriptor.FIELDNAMES[0], HASH_UQ, txId);
-      this.createIndexWithoutSchemaName(MediaSegmentDescriptor.ENTITY, MediaSegmentDescriptor.FIELDNAMES[1], IndexType.HASH, txId);
+      this.createIndex(entityName, MediaSegmentDescriptor.FIELDNAMES[0], IndexType.BTREE_UQ, txId);
```
silvanheller, line 144: Is `HASH` now simply `BTREE` or are there also implementation changes in Cottontail?

cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
```diff
@@ -358,20 +341,9 @@ public static Type mapAttributeType(AttributeDefinition.AttributeType type) {
...
-      this.cottontail.client.create(index);
-    }catch(StatusRuntimeException e){
-      if(e.getStatus().getCode()== Status.ALREADY_EXISTS.getCode()){
-        LOGGER.warn("Index {} was not created because it already exists", indexName);
-        return false;
-      }
-      throw e;
-    }
-    return true;
+    this.cottontail.client.create(index);
```
silvanheller, line 347: I don't understand the diff here, I think something went wrong while merging. The functionality to create an Index for the ID seems to have been deleted, and the handling for if an index already exists seems also to have been deleted. Is this a merge-error (if so, I can fix that) or is this intentional?

silvanheller (changes requested):
There are some changes which require additional documentation, and some which might be merge errors. I think it would be good to open also a PR for vitrivr-ng to `dev`, which we can have a look at in parallel.

cineast-api/src/main/java/org/vitrivr/cineast/api/messages/query/TemporalQuery.java
lucaro: I didn't touch vitrivr-ng at all, just used the branch that has the pose query interface. Without this setter, it didn't work. That branch might not be completely current though, I'll have to check.

cineast-api/src/main/java/org/vitrivr/cineast/api/rest/resolvers/FileSystemThumbnailResolver.java
lucaro: This might be able to be removed again. It was needed to be compatible with the format of the v3c dataset. Maybe something like this should be more easily configurable.

cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
lucaro: This is a cottontail change, xodus comes with new index types. Maybe @ppanopticon can elaborate?

cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
lucaro: This looks like a merge error to me as well, git apparently got confused with the change order when I resurrected the dev branch.

cineast-api/src/main/java/org/vitrivr/cineast/api/messages/query/TemporalQuery.java
lucaro: With the most recent version, it's no longer needed, so I removed it.

cineast-core/src/main/java/org/vitrivr/cineast/core/db/cottontaildb/CottontailEntityCreator.java
ppanopticon: This is simply a rename. 

What used to be called HASH and HASH_UQ were Tree based indexes all along (even for MapDB). I just took the opportunity to reflect that in the naming of the different enumerations.

silvanheller (commented):
Looks good to me. I'll have to play around with the UI & import to see if it works, but generally speaking this works for me.

ppanopticon (changes requested):
Tests still fail because the branch requires 0.15.0-SNAPSHOT version of Cottontail DB to run, which is not available as Docker Image yet.

We can finalise this PR, once Cottontail DB has been released.

silvanheller (approved):
LGTM! We'll wait with merging until v0.15 of Cottontail is out, but this looks really cool.

ppanopticon (approved):

PR #294, opened 2022-05-05
Upgrade/tensorflow

Spiess:
Upgrades the used version of TensorFlow from 0.3.1 to 0.4.1.

## Persistent issues
- The `MovenetMultiposeDetector` causes a NullPointerException at `Skeleton.java:96`, but this issue also occurs on master.

PR #296, opened 2022-05-05
Feature/TensorFlow GPU

### LINKED ISSUE #293: TensorFlow GPU support
- Add GPU support through the Java TensorFlow GPU artifacts.
- Add build flag to differentiate between CPU and GPU builds.

Spiess:
Implements additional build flag `-PGPU=true` to use GPU version of TensorFlow.

Closes #293.

Will be made concrete once #294 is merged.


silvanheller:
Have you tested that this is actually just a drop-in replacement and uses the GPU (e.g. using `nvidia-smi`)

Spiess:
Please excuse, I forgot to mention that I did test this and could determine that the GPU is used, but since different item extractions are not batched, this could still be improved in the future.

silvanheller (commented):
I tested this on ubuntu and confirm it works. There is however not really any noticeable speedup for CLIP. One request: Could the jar have a flag indicating whether it was built with gpu support?

lucaro:
> I tested this on ubuntu and can confirm it works. There is however not really any noticeable speedup for CLIP. One request: Could the jar have a flag indicating whether it was built with gpu support?

I think that would also make sense for the application build. Alternatively, we might just add a CLI command that lists the currently available GPU compute resources. That would also serve as a debug feature.

lucaro:
I just tried what happens if you include `tensorflow-core-platform-gpu` on a machine without cuda support. It tries to load cuda, fails, and falls back to the CPU. Apart from some additional error messages in the log and a slightly larger size of the dependencies, there is, therefore, no downside in always loading the gpu-enabled version. The switch is not even necessary. The same behavior can be observed with `tensorflow-core-platform-mkl`. Interestingly enough, `tensorflow-core-platform-mkl-gpu` which contains both the intel and the nvidia accelerator support did not work on my machine, otherwise I'd have suggested that as a default.

lucaro:
I removed the switch and set it directly to use the GPU-enabled dependency. We can merge this now.

lucaro (approved):

PR #297, opened 2022-05-07
Clip upgrades

silvanheller:
Contingent on https://github.com/vitrivr/cottontaildb/issues/117 being merged, this PR:
- Makes CLIP fully functional with sensible correspondence functions, tested on the LSC dataset (it assumes that the cosine distance is an actual distance and not a similarity function)
- Fixes a bug where thumbnails would not be resolved if there is a `_` in the objectid (e.g. `i_coffee_cup`)
- Explicitly makes sure that the preprocessing is the same as in the CLIP python examples
- Implements farthest-neighbor search on the `DBSelector` if ever there is a distance-function which is actually a similarity-function
- Adds a correspondence function for this case

Example:
image


lucaro (approved):
Since https://github.com/vitrivr/cottontaildb/pull/118 is already merged, this should now be good to go as well.

PR #305, opened 2022-05-13
Fixed #304 by using a new version

### LINKED ISSUE #304: No releases for v3.11.3
Due to a bug which was fixed in #300 there is no release for that version.

sauterl:


lucaro (approved):
Do we really need 4-element version numbers?

sauterl:
Closed this PR since its only purpose was to re-generate the release artifacts of #298 (v3.11.3).
The effects were produced by simply having a corresponding tag (without actually having that version number).

With this, we now have created far more conversation and documentation trails than really neded and we do have new artifacts.
For the future we shall stick to three element version numbers, as pointed out above.

Issue #306, opened 2022-05-13
Standalone does not show a help message for <command> -h


sauterl:
Version: v3.11.+
Standalone Mode (i.e. obtained by `./gradlew distZip` then using the dist of module runtime)
OS: Tested with ubuntu 20.04 and windows 10
Command: every command, for instance `./cineast-runtime-3.11.3/bin/cineast-runtime cineast.json extract -h`

Currently, if a command is executed with the commonly expected POSIX help option `-h`, an exception is thrown without further notice.

While the exception mentions what went wrong, this is not a desirabled behaviour. Ideally, the help message for that command is shown and optionally what went wrong with this particular call.

PR #307, opened 2022-05-13
Standalone commands / CLI respect -h / --help for help

sauterl:
Addresses #306 by slightly refactoring the current command hierarchy and introducing the following behaviour:

When the `-h` (or `--help`) option is given for a command (i.e. `<dist> path/to/config.json <command> -h`),
the missing required options are listed and then the help message is shown on std err.

For the interactive CLI, the same logic applies, however the information is printed on std out with the colour red.

**Important: This PR already contains the next version bump**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ allprojects {
     group = 'org.vitrivr'
 
     /* Our current version, on dev branch this should always be release+1-SNAPSHOT */
-    version = '3.11.3'
+    version = '3.11.4'
 
     apply plugin: 'java-library'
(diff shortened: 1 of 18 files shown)


silvanheller (approved):
LGTM, i applied the codestyle in the repo (https://github.com/vitrivr/cineast/blob/master/docs/intellij-java-google-style.xml) to the changes

sauterl:
Thanks! This is weird though, I do have the proper style guide in my IDE activated - must've forgotten to reformat all touched files...

PR #309, opened 2022-05-16
Optimize CLIP model loading

Spiess:
Optimizes loading the CLIP model by only loading it when needed and keeping it loaded between segments.


lucaro (approved):
👍

PR #315, opened 2022-05-17
Feature/result caching

lucaro:
Adds a simple caching mechanism, so repeated queries don't need to be re-evaluated.


silvanheller (approved):
LGTM!

PR #321, opened 2022-06-21
LSC22

silvanheller:
- Fixing distinct element rest request handler crashing when receiving a NothingProvider from Cottontail (which is in turn caused by null being returned)
- Improving logging for NothingProvider, as this should basically never happen


cineast-core/src/main/java/org/vitrivr/cineast/core/data/providers/primitive/NothingProvider.java
```diff
@@ -4,6 +4,8 @@ public class NothingProvider implements PrimitiveTypeProvider {
 
   public static final NothingProvider INSTANCE = new NothingProvider();
 
+  public NothingProvider() {}
```
lucaro, line 7: Is there a specific reason you need more instances of that one? We mainly used it as a singleton so far, since all of them are the same anyway.

lucaro (approved):

PR #323, opened 2022-07-23
Fix for ClipTokenizer

lucaro:


silvanheller (approved):
Works perfectly, thanks for the work investigating the issue and the fix!

PR #324, opened 2022-07-25
Making decorator for text retrieval configurable

silvanheller:
Some time ago, each term used to be decorated with `~` at the end for fuzzy matches. This got lost at some point, which leads to a query for `JAMES` to not match `AMES` in an OCR column. To fix this, this PR adds configurable decorators for all text-retrieval classes. Default behavior stays the same.

Additionally, it adds the option to configure entity-names for text retrieval classes and fixes rare edge-cases where batch-import would cause problems.

### DIFF
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/features/ProvidedOcrSearch.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/features/ProvidedOcrSearch.java
@@ -1,5 +1,6 @@
 package org.vitrivr.cineast.core.features;
 
+import java.util.LinkedHashMap;
 import org.vitrivr.cineast.core.features.abstracts.AbstractTextRetriever;
 
 /**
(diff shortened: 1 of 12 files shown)


cineast-core/src/main/java/org/vitrivr/cineast/core/features/ProvidedOcrSearch.java
```diff
@@ -15,4 +16,9 @@ public class ProvidedOcrSearch extends AbstractTextRetriever {
   public ProvidedOcrSearch() {
     super(ProvidedOcrSearch.PROVIDED_OCR_SEARCH_TABLE_NAME);
   }
+
+  public ProvidedOcrSearch(LinkedHashMap<String, String> properties) {
```
lucaro, line 20: Why specifically does this need to be a `LinkedHashMap` and not just a `Map`?

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractTextRetriever.java
```diff
@@ -62,7 +69,19 @@ public abstract class AbstractTextRetriever implements Retriever, Extractor {
    * @param tableName Name of the table/entity used to store the data
    */
   public AbstractTextRetriever(String tableName) {
-    this.tableName = tableName;
+    this(tableName, new LinkedHashMap<>());
+  }
+
+  public AbstractTextRetriever(String defaultTableName, LinkedHashMap<String, String> properties) {
```
lucaro, line 75: Again, why so specific?

lucaro (changes requested):

silvanheller:
I've made changes to all features which were using a specific type of map in their constructor and did not need it.

lucaro (approved):

PR #325, opened 2022-07-26
Implemented OS switch so macOS builds use CPU version of TensorFlow.

### LINKED ISSUE #313: No tensorflow-core-platform-gpu artifacts for macOS
There are no `tensorflow-core-platform-gpu` artifacts for macOS because TensorFlow does not provide GPU binaries for macOS.
While Cineast builds with `tensorflow-core-platform-gpu` behave the same on linux systems without GPU as builds with `tensorflow-core-platform`, they do not work at all on macOS.

A possible solution would be to introduce a build flag for the CPU version.

Spiess:
Fixes #313.

Tested to work as expected with builds on macOS and Ubuntu.

### DIFF
--- a/cineast-core/build.gradle
+++ b/cineast-core/build.gradle
@@ -1,3 +1,5 @@
+import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform
+
 repositories {
     maven {
         url "https://oss.sonatype.org/content/repositories/snapshots/"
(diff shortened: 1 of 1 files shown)


lucaro (commented):
Is this a problem for MacOS in general or only for the ARM-based machines?

Spiess:
There are no TensorFlow binaries for macOS ARM at all, this is only for macOS x64.

lucaro (approved):

PR #329, opened 2022-07-26
Visual-Text Co-Embedding image retrieval

Spiess:
Previously the visual-text co-embedding only allowed retrieval with text terms and would even retrieve using the empty string provided by default if no text term was specified.
This has been fixed with appropriate error messages and returns.
Retrieval using image terms was implemented.


silvanheller (approved):
LGTM

Issue #330, opened 2022-07-26
TFInvalidArgumentException for CLIP


silvanheller:
When running a significant amount of queries on Cineast, you start running into CLIP issues (see error log below). I have not yet found a way to trigger this error deterministically (the same query will trigger it sometimes and other times not), and it is sometimes solved after a clean & rebuild.

I'm filing this here for future reference.

```
2022-07-26 15:13:35.515 [ssq-0] WARN o.v.c.s.r.ContinuousQueryDispatcher - java.util.concurrent.ExecutionException: org.tensorflow.exceptions.TFInvalidArgumentException: assertion failed: [Gather indices are out of bounds, please double check the indices and retry.] [Condition x == y did not hold element-wise:] [x (LogicalAnd:0) = ] [[0 1 1...]...] [y (assert_equal_1/y:0) = ] [1]
         [[{{function_node assert_equal_1_Assert_AssertGuard_false_274}}{{node assert_equal_1/Assert/AssertGuard/Assert}}]]
        at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
        at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
        at org.vitrivr.cineast.standalone.runtime.ContinuousQueryDispatcher.extractResults(ContinuousQueryDispatcher.java:188)
        at org.vitrivr.cineast.standalone.runtime.ContinuousQueryDispatcher.doRetrieve(ContinuousQueryDispatcher.java:157)
        at org.vitrivr.cineast.standalone.runtime.ContinuousQueryDispatcher.retrieve(ContinuousQueryDispatcher.java:98)
        at org.vitrivr.cineast.standalone.util.ContinuousRetrievalLogic.retrieve(ContinuousRetrievalLogic.java:40)
        at org.vitrivr.cineast.api.websocket.handlers.queries.TemporalQueryMessageHandler.lambda$execute$9(TemporalQueryMessageHandler.java:120)
        at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: org.tensorflow.exceptions.TFInvalidArgumentException: assertion failed: [Gather indices are out of bounds, please double check the indices and retry.] [Condition x == y did not hold element-wise:] [x (LogicalAnd:0) = ] [[0 1 1...]...] [y (assert_equal_1/y:0) = ] [1]
[...]

PR #332, opened 2022-07-27
Fixing CLIP Parallelism Issues

silvanheller:
This PR bumps the Tensorflow version to 0.5.0-SNAPSHOT and solves parallel calls to to the CLIP model using the apache-commons poling library. This is necessary, testing has shown that removing synchronization leads to the issue described in https://github.com/vitrivr/cineast/issues/330 re-appearing.

This also fixes https://github.com/vitrivr/cineast/issues/330. 

This PR is waiting until 0.5.0 is stable.


lucaro:
I don't understand the intent of this. Why parallelize at all? TensorFlow will make use of as many resources as it can, so presumably, trying to run multiple instances of the same model in parallel on the same hardware will conflict and offer little to no benefit over sequential evaluation. Or am I mistaken in that regard?

PR #333, opened 2022-08-08
Replaced potentially problematic libraries

lucaro:


cineast-core/build.gradle
```diff
@@ -90,18 +88,15 @@ dependencies {
...
     /** Mockito for unit testing */
-    implementation 'org.mockito:mockito-inline:4.5.1'
+    implementation 'org.mockito:mockito-inline:4.6.1'
 
     /** Tensorflow (Java). */
-    if (DefaultNativePlatform.currentOperatingSystem.isMacOsX())
-        api group: "org.tensorflow", name: "tensorflow-core-platform", version: version_tensorflow
-    else
-        api group: "org.tensorflow", name: "tensorflow-core-platform-gpu", version: version_tensorflow
+    api group: "org.tensorflow", name: "tensorflow-core-platform-gpu", version: version_tensorflow
```
silvanheller, line 99: Is there a reason we're reverting to not having macOS specific bindings? https://github.com/vitrivr/cineast/pull/325

silvanheller (commented):
Thanks for the refactoring!

cineast-core/build.gradle
lucaro: No, that was not intentional. Had that change already been merged? Might be a git merge issue on my end, I will have a look.

silvanheller (approved):

PR #331, opened 2022-07-27
Fixing CLIP issues for parallel execution 

silvanheller:
This PR bumps the Tensorflow version to 0.5.0-SNAPSHOT and prevents parallel calls to the CLIP model. This is necessary, testing has shown that removing synchronization leads to the issue described in #330 re-appearing.

While this leads to performance degradations in parallel evaluations, it fixes #330. For the future, we might want to load the model multiple times and have some sort of sharing for parallel CLIP queries.

PR #334, opened 2022-08-11
Fix Parallel Execution of CLIP Feature

silvanheller:
This PR solves parallel calls to to the CLIP feature (#330), which were caused by initialization problems of CLIP & the tokenizer in heavily parallel scenarios.

It does not prevent parallel model-calls, nor parallel execution of the feature.

As the new Tensorflow version is still on SNAPSHOT, this PR remains on 0.4.1 and is thus ready for review as a slightly updated version of #331 #332 

Functionality is fully tested with large-scale evaluation runs for parallel model execution.


Spiess (approved):
I do not have the appropriate setup to test the parallel execution of the CLIP Feature, but I was able to test that all touched features still work.

PR #339, opened 2022-10-03
Rework extraction API endpoints

simonpeterhans:
This PR fixes the broken REST API endpoints for feature extraction from items.

The session concept is roughly kept, meaning that an extraction session first has to be opened before items can be added, and should be closed once all items have been added.

The path to the extraction configuration to use can be specified in the config file with `sessionExtractionConfigLocation`.

The `ExtractionContextProvider` interface is used in `GenericExtractionItemHandler` instead of an `IngestConfig` to properly reuse the handler.

___

Since I will not have any time for further changes or contributions to this, @silvanheller asked me to open this PR as-is -- feel very free to make further changes to the respective branch.


silvanheller (approved):
LGTM, thanks for the work on fixing the endpoint! I've added a check for the config flag to allow / disallow extraction

PR #343, opened 2022-10-13
Switch from deprecated openjdk to eclipse-temurin base

x4e-jonas:
See notice on https://hub.docker.com/_/openjdk

Alternatively you could also use the `gradle:jdk17` image to avoid downloading gradle on every build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,11 @@
-FROM openjdk:17-buster AS build
+FROM gradle:jdk17 AS build
 
-RUN apt-get update && \
-  apt-get install -y maven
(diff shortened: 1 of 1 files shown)


silvanheller (dismissed):
Thanks for the PR! I don't have a strong preference on the jdk we use, i'll leave it up to @lucaro if eclipse-temurin is the way to go.

lucaro:
When we're already changing the base image, I'd just go for `gradle:jdk17`.

silvanheller (approved):
Thanks for the PR!

silvanheller (approved):

Issue #342, opened 2022-10-13
`data.vitrivr.org` server misconfigured


x4e-jonas:
When using HTTPS, `data.vitrivr.org` points to a completely different virtual host:
- http://data.vitrivr.org/DeepLab/cityscapes.pb
- https://data.vitrivr.org/DeepLab/cityscapes.pb

silvanheller:
Thanks for pointing this out, it should be fixed now.

PR #346, opened 2022-10-13
Use HTTPS for externalFiles

x4e-jonas:
#342 follow-up


silvanheller (approved):

lucaro:
I don't see the benefit of enforcing this computational overhead.

PR #350, opened 2022-10-17
Fixing CLI not exiting properly

### LINKED ISSUE #349: Exit and ^D no longer exit the CLI
The updated CLI main (#345) no longer exits with the `exit` keyword or `^D` end-of-file character.

This is presumably due to the API shutdowns having been moved to a shutdown hook, which is not triggered, since the APIs it shuts down prevent the VM from shutting down.

silvanheller:
This PR fixes #349. It does not interfere with the aim of #345, as there is a while(true) loop if the CLI is not enabled.

### DIFF
--- a/cineast-api/src/main/java/org/vitrivr/cineast/api/Main.java
+++ b/cineast-api/src/main/java/org/vitrivr/cineast/api/Main.java
@@ -69,5 +69,6 @@ public class Main {
     } catch (InterruptedException e) {
       e.printStackTrace();
     }
+    System.exit(0);
   }
 }


Spiess (approved):

lucaro:
Does this actually trigger the shutdown hook to close the API properly or does it just kill the process?
Wouldn't it be cleaner to only register the shutdown hook when the CLI is not enabled and otherwise handle the shutdown mechanism as before?

Spiess:
I have tested this with the CLI enabled and it produces the log output from the shutdown hook, so it appears to be closed correctly as far as I can tell.

PR #352, opened 2022-11-03
Added ConstantLengthVideoSegmenter

lucaro:


cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/video/ConstantLengthVideoSegmenter.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Override
+  public boolean complete() {
+    return this.outputQueue.isEmpty() && this.complete.get();
+  }
+
+  @Override
+  public void close() {
+
```
silvanheller, line 80: Does this on purpose not interfere with the loop in the `run` method?

lucaro: Nope, that was not intentional. Fixed now.

silvanheller (approved):
LGTM

PR #356, opened 2022-11-16
Fixed #353 by respecting config value for segmenter

### LINKED ISSUE #353: Segmenter config ignored during extraction
The `IngestConfig` allows for specification of a specific `Segmenter`. This is however not respected when setting up the extraction pipeline.

sauterl:
Previously, the GenericExtractionItemHandler did not respect the config-provided segmenter, which is being fixed by this PR.

If no segmenter is specified, reasonable defaults are used per media type, this logic was not changed.

### DIFF
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/general/Segmenter.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/general/Segmenter.java
@@ -1,5 +1,6 @@
 package org.vitrivr.cineast.core.extraction.segmenter.general;
 
+import org.vitrivr.cineast.core.data.MediaType;
 import org.vitrivr.cineast.core.data.entities.MediaObjectDescriptor;
 import org.vitrivr.cineast.core.data.segments.SegmentContainer;
 import org.vitrivr.cineast.core.extraction.decode.general.Decoder;
@@ -39,4 +40,10 @@ public interface Segmenter<A> extends Runnable, AutoCloseable {
    */
   @Override
   void close();
+
+  /**
+   * Specifies for which media type this segmenter is
+   * @return The {@link MediaType} this {@link Segmenter} is for.
+   */
+  MediaType getMediaType();
 }
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/general/Segmenter.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/general/Segmenter.java
@@ -5,6 +5,8 @@ import org.vitrivr.cineast.core.data.entities.MediaObjectDescriptor;
 import org.vitrivr.cineast.core.data.segments.SegmentContainer;
 import org.vitrivr.cineast.core.extraction.decode.general.Decoder;
 
+import java.util.Set;
+
 /**
  * {@link Segmenter}s split a media file into chunks (segments). The nature of that chunk is specific to the media type and the{@link Segmenter}s implementation. A segment could be anything from a shot of a video to an arbitrary part of song or just a single image.
  */
@@ -42,8 +44,8 @@ public interface Segmenter<A> extends Runnable, AutoCloseable {
   void close();
 
   /**
-   * Specifies for which media type this segmenter is
-   * @return The {@link MediaType} this {@link Segmenter} is for.
+   * Specifies for which media types this segmenter is
+   * @return The set of {@link MediaType}s this {@link Segmenter} is for.
    */
-  MediaType getMediaType();
+  Set<MediaType> getMediaTypes();
 }
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/image/ImageSequenceSegmenter.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/image/ImageSequenceSegmenter.java
@@ -132,7 +132,7 @@ public class ImageSequenceSegmenter implements Segmenter<ImageSequence> {
   }
 
   /**
-   * Returns {@link MediaType#IMAGE}, as this {@link Segmenter} is for images
+   * Returns {@link MediaType#IMAGE_SEQUENCE}, as this {@link Segmenter} is for image sequences
    * @return
    */
   @Override
--- a/cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/image/ImageSequenceSegmenter.java
+++ b/cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/image/ImageSequenceSegmenter.java
@@ -127,4 +127,13 @@ public class ImageSequenceSegmenter implements Segmenter<ImageSequence> {
       this.running = false;
     }
   }
+
+  /**
+   * Returns {@link MediaType#IMAGE}, as this {@link Segmenter} is for images
+   * @return
+   */
+  @Override
+  public MediaType getMediaType() {
+    return MediaType.IMAGE;
+  }
 }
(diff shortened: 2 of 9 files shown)


cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/general/Segmenter.java
```diff
@@ -39,4 +40,10 @@
    */
   @Override
   void close();
+
+  /**
+   * Specifies for which media type this segmenter is
+   * @return The {@link MediaType} this {@link Segmenter} is for.
+   */
+  MediaType getMediaType();
```
lucaro, line 48: I'd argue that a segmenter can work with more than only one media type and therefore, this should be a collection, ideally a set.

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
```diff
@@ -119,15 +119,24 @@ public GenericExtractionItemHandler(ExtractionContainerProvider pathProvider, Ex
...
-    handlers.put(MediaType.MODEL3D, new ImmutablePair<>(ModularMeshDecoder::new, () -> new PassthroughSegmenter<Mesh>() {
+    // #353: Respect the given segmenter
+    final Segmenter segmenter = context.newSegmenter();
+    final MediaType segmenterType = segmenter.getMediaType();
+
+    handlers.put(MediaType.IMAGE, new ImmutablePair<>(DefaultImageDecoder::new, () -> segmenterType == MediaType.IMAGE ? segmenter : new ImageSegmenter(context)));
+    handlers.put(MediaType.IMAGE_SEQUENCE, new ImmutablePair<>(ImageSequenceDecoder::new, () -> segmenterType == MediaType.IMAGE_SEQUENCE ? segmenter : new ImageSequenceSegmenter(context)));
+    handlers.put(MediaType.AUDIO, new ImmutablePair<>(FFMpegAudioDecoder::new, () -> segmenterType == MediaType.AUDIO ? segmenter : new ConstantLengthAudioSegmenter(context)));
+    handlers.put(MediaType.VIDEO, new ImmutablePair<>(FFMpegVideoDecoder::new, () -> segmenterType == MediaType.VIDEO ? segmenter : new VideoHistogramSegmenter(context)));
+    handlers.put(MediaType.MODEL3D, new ImmutablePair<>(ModularMeshDecoder::new, () -> segmenterType == MediaType.MODEL3D ? segmenter : new PassthroughSegmenter<Mesh>() {
```
lucaro, line 130: This is needlessly verbose. Simply add all the defaults to the map first and then overwrite the values with the ones specified in the config.

lucaro (changes requested):

cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/general/Segmenter.java
sauterl: That is a valid point. However, from my point of view it would be an interesting approach to derive the media type (by some registry?) from the actual type argument of the segmenter.
For now, I'm happy to make this change!

lucaro: I don't see how you would want to derive this, as it is dependent on the actual logic of the segmenter. The most trivial case would be the segmenter that just takes the entire object as one segment, that works for any type. Since we only have a small number of these segmenters anyway, it's easier for them to declare what types they support rather than trying to build something elaborate and possibly error prone so they don't have to.

lucaro (approved):
Looks reasonable

cineast-core/src/main/java/org/vitrivr/cineast/core/extraction/segmenter/image/ImageSequenceSegmenter.java
```diff
@@ -127,4 +130,13 @@ public void run() {
...
     }
   }
+
+  /**
+   * Returns {@link MediaType#IMAGE}, as this {@link Segmenter} is for images
+   * @return
+   */
+  @Override
+  public Set<MediaType> getMediaTypes() {
+    return Sets.newHashSet(MediaType.IMAGE);
```
silvanheller, line 140: Shouldn't this be `MediaType.IMAGE_SEQUENCE`?

silvanheller (changes requested):

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
```diff
@@ -118,21 +116,38 @@ public GenericExtractionItemHandler(ExtractionContainerProvider pathProvider, Ex
...
+    handlers.put(MediaType.VIDEO, new ImmutablePair<>(FFMpegVideoDecoder::new, () ->  new VideoHistogramSegmenter(context)));
+    handlers.put(MediaType.MODEL3D, new ImmutablePair<>(ModularMeshDecoder::new, () ->  new PassthroughSegmenter<Mesh>() {
       @Override
       protected SegmentContainer getSegmentFromContent(Mesh content) {
         return new Model3DSegment(content);
       }
+
+      @Override
+      public Set<MediaType> getMediaTypes() {
+        return Sets.newHashSet(MediaType.MODEL3D);
```
silvanheller, line 134: And I don't understand why this is `MODEL3D`

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
```diff
@@ -118,21 +116,38 @@ public GenericExtractionItemHandler(ExtractionContainerProvider pathProvider, Ex
...
+      }
     }));
+    // #353: Respect the given segmenter
```
silvanheller, line 137: Additionally, i don't think referencing issue numbers in comments is sensible

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
lucaro: Since it is supposed to be generic, it should either have all the types or not implement the method, I guess.

cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
```diff
@@ -118,21 +116,38 @@ public GenericExtractionItemHandler(ExtractionContainerProvider pathProvider, Ex
...
+        handlers.put(t, new ImmutablePair<>(handlers.get(t).getLeft(), () -> segmenter));
+      });
+    }
+
     //Config overwrite
     Config.sharedConfig().getDecoders().forEach((type, decoderConfig) -> {
       handlers.put(type, new ImmutablePair<>(ReflectionHelper.newDecoder(decoderConfig.getDecoder(), type), handlers.getOrDefault(type, null)).getRight());
     });
+
     //TODO Config should allow for multiple segmenters
```
silvanheller, line 151: Is this TODO adressed with this PR?

lucaro: By my understanding no, as this refers to the option to specify multiple segmenters in the config, presumably for different media types.

sauterl: Exactly, this PR only tackles #353 and does not do anything else

lucaro (approved):

silvanheller (approved):

PR #358, opened 2022-11-21
Fixed segmenter config not usable due to broken code

sauterl:
With the current version on main, the segmenter is being closed upon initialisation. Obviously, this is not very helpful and leads to the extraction being unable to start.


lucaro (approved):
Ah, yes, I didn't notice the `try` block, that indeed should not be there...

PR #360, opened 2022-11-29
Feature/parent folder name object id generator

net-cscience-raphael:
Created a new ObjectIdGenerator which uses the name of the containing parent folder for the object id.
With "prefix": "true" the Generator includes type prefix.


net-cscience-raphael:
To use the "ParentFolderNameObjectIdGenerator"  set configuration in extraction_config.json as follows:

```json
    "id": {
      "name": "ParentFolderNameObjectIdGenerator",
      "properties": {
        "prefix": "true"
      }
    }
```

lucaro (commented):
What is the expected behavior of this generator supposed to be in cases where a directory contains more than one relevant media file?

silvanheller (commented):
Thanks for the PR! I noticed that our existing ID generators have some suboptimal code, so I refactored them in #361. This should require minimal changes to your PR, you can mostly just take the changes from the `FileNameObjectIdGenerator`.

Spiess (approved):

PR #361, opened 2022-12-01
Refactoring Object Id Generation

silvanheller:
Motivated by my review #360, I refactored and cleaned up the existing object id generators.

- Using `Optional` instead of "null" as a string (!)
- Removed unused code
- Using `getOrDefault` instead of more verbose code
- Throwing exceptions instead of generating objects with `null` as id.


Spiess (approved):
Thanks for the refactoring, I find it to be much more readable.

PR #362, opened 2023-01-24
Feature/ffmpeg

lucaro:
Added decoder based on external ffmpeg process. Tested and confirmed to be working with video, audio data might be unreliable depending on input due to lack of resampling. This might be helpful for extractions where no audio is needed and the decoder based on libffmpeg leaks memory.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform
+
 plugins {
     id 'com.github.johnrengelman.shadow' version '7.1.2'
     id 'de.undercouch.download' version "5.0.5"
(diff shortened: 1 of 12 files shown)


silvanheller (commented):
I like it! A few thoughts:
- I'll let @sauterl comment on whether this works for long-running extractions
- We'll need to update the documentation to make sure people execute the appropriate task

lucaro:
Currently, there also isn't any explicit mechanism to select it in the config, so using it currently requires a code change. The responsible part of the config logic was recently adjusted anyway and I think there are still some issues with it. It might be worth discussing how much flexibility we want to expose in this case and what should just be handled using automatisms and 'reasonable defaults'.

PR #364, opened 2023-02-14
Improved error message when no config is available

silvanheller:
Based on #363, if no config path is provided & no config file is available, the behavior is inconsistent in that Cineast does not shut down but proceeds.

This change makes sure the behavior is consistent


lucaro (commented):
Looks reasonable to me. While you are at it, the behavior is almost certainly also inconsistent in the runtime entry point.

lucaro (approved):

PR #365, opened 2023-02-26
Fix/extraction segmenter config

lucaro:
FIxes error during extration initialization in case no segmenter is specified in the extraction config.

### DIFF
--- a/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
+++ b/cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
@@ -139,10 +139,12 @@ public class GenericExtractionItemHandler implements Runnable, ExtractionItemPro
     // #353: Respect the given segmenter
     final Set<MediaType> segmenterTypes;
     final Segmenter<Object> segmenter = context.newSegmenter();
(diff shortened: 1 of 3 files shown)


cineast-runtime/src/main/java/org/vitrivr/cineast/standalone/run/GenericExtractionItemHandler.java
```diff
@@ -139,10 +139,12 @@ public Set<MediaType> getMediaTypes() {
     // #353: Respect the given segmenter
     final Set<MediaType> segmenterTypes;
     final Segmenter<Object> segmenter = context.newSegmenter();
-    segmenterTypes = segmenter.getMediaTypes();
-    segmenterTypes.forEach(t -> {
-      handlers.put(t, new ImmutablePair<>(handlers.get(t).getLeft(), () -> segmenter));
-    });
+    if (segmenter != null) {
```
silvanheller, line 142: Can we add a log-message if the segmenter is null? I don't see the scenario where the segmenter is null but we have a proper config

lucaro: The segmenter is optional in the config, so there are plenty of scenarios where this would be the case. But sure, we can log this.

Spiess (approved):

PR #366, opened 2023-03-07
Cleanup & bugfixes for default setup

silvanheller:
Minor bugfixes to ensure a smooth process for users who try to extract & retrieve data with the default config & README


lucaro (approved):
looks reasonable to me

Spiess (approved):

PR #373, opened 2023-04-28
DBSelector Documentation Updates

silvanheller:
This PR unifies variable names & cleans up documentation in the `DBSelector` in preparation on upcoming work for multi-collection support.

Specifically, it adds query-configs to most retrieval methods where they were not already present, adds SQL-Like documentation to methods, cleans up unused code, makes clear which methods are convenience methods and which can be overriden by underlying implementations for performance-gains.


lucaro (approved):
👍

PR #369, opened 2023-04-24
Cross modal Retrevial of 3D models in vitrivr

net-cscience-raphael:
### Overview
This pull request incorporates the code from the bachelor thesis titled "Cross-modal Retrieval of 3D Models in vitrivr." (See. [Thesis](http://archiv.cs.unibas.ch/2023/waltenspul_raphael/index.html.))

The primary contributions of this thesis involve extracting feature vectors using visual-text co-embedding. For that reason the renderer implementation, which previously relied on the Java binding for the OpenGL API (JOGL), has been replaced with a new renderer based on the Lightweight Java Game Library (LWJGL).

While the original JOGL renderer remains in the project, its interface is annotated as _Deprecated_. The new LWJGL renderer is located in the package `org/vitrivr/cineast/core/render/lwjgl`, with shader programs found in `resources/renderer/lwjgl/shaders`.

The renderer package is organized into the following sub-packages:
- `.engine`: The main class for the rendering engine, which includes the window, scene, and render object, providing rendering functionality.
- `.glmodel`: These wrapper classes enable generic models to be used in the OpenGL context.
- `.render`: This sub-package manages shader program loading and compilation, as well as 3D model rendering using the LWJGL engine.
- `.renderer`: This package implementing the Cineast-specific rendering functions.
- `.scene`: These classes represent all 3D scene elements, such as models, cameras, and projections, enabling OpenGL context rendering.
- `.util`: This package provides utility classes, including a finite state machine for controlling the rendering process.
- `.window`: This represents the 2D projection of the 3D model, generating the light field from the data in this window. Both headless and visible rendering are supported.

Additionally, a new generic 3D model implementation is introduced in the package `org/vitrivr/cineast/core/data/m3d/texturemodel`. The <a href="https://javadoc.lwjgl.org/org/lwjgl/assimp/Assimp.html">Assimp</a> library was utilized to load the 3d models.

The previous "Mesh" model is retained in the project for now. In subsequent steps, it is planned to migrate the functionality of the previous "Mesh" model to the new generic 3D model.

The package 'org/vitrivr/cineast/core/util/texturemodel' contains methods for determining the optimal viewpoint / camera position.

The new LWJGL renderer is implemented in the following classes:
- `org/vitrivr/cineast/core/features/Lightfield.java`
- `org/vitrivr/cineast/core/features/VisualTextCoEmbedding.java`
- `org/vitrivr/cineast/core/features/exporter/Model3DThumbnailExporter.java`

PR #375, opened 2023-05-04
Extraction & QoL Improvements

silvanheller:
This PR upgrades contains extraction-improvements.
- Upgrading TF-java to v0.5.0 (needed for Apple Silicon) done by @Spiess 
- Clarifications & improved defaults for extraction
- Upgrading OpenCV to 4.7.0 (needed for Apple Silicon)
- Bumping FFMPEG version to avoid javacpp version conflicts
- Fixing error during extraction introduced in #373 
- Fixing query id generation for cottontail as a followup to #373 
- Adding a config flag for the RenderWorker because #369 means that there are stacktraces when vitrivr starts otherwise
- Removing duplicate logging of log level introduced in #369 
- Consistently applying `ID` vs `Id` spelling in variable names


lucaro (approved):
👍

Issue #377, opened 2023-05-10
Make log level configurable


x4e-jonas:
By default, Cineast runs in DEBUG mode. This is nut suitable for production deployments.

The log level should be runtime configurable (either by argument, option in config file or environment variable) and it might be a good idea to consider and document a sane default value.

x4e-jonas:
Fixed by #378

PR #378, opened 2023-05-10
Make log level configurable (#377)

x4e-jonas:
Uses `LOG_LEVEL` to configure log level. Default changed to `INFO`.


silvanheller (approved):

silvanheller (approved):

lucaro (changes requested):
The way this is implemented only makes sense under the assumption that cineast is the only application running in a particular environment, such as a docker container. In all other cases, it uses a generic environment variable without any context. To make this more generally applicable, I'd suggest calling the variable something like `CINEAST_LOG_LEVEL`, in case there are several applications in the same environment that use a similar mechanism.

lucaro (approved):
Looks reasonable to me

Issue #370, opened 2023-04-28
Configurable Feature Implementation to Access External Retrieval Features


silvanheller:
Based on the work done in the [VBS 23 Branch](https://github.com/vitrivr/cineast/tree/VBS23), and to make vitrivr usable with retrieval features that are either not easily portable to java or only accessible externally, we should integrate a configurable retrieval feature which calls for extraction and retrieval an external REST API.

The feature should have at least the following configuration options:
- API endpoint
- Type: Vector or Text
- Vector length

To keep the repo clean, we do not provide an example external feature (e.g., python) in the repo itself but in the wiki.

lucaro:
Doing this properly would presumably require some specified API with which these external services can be queried.

sauterl:
Also, I would like to add that in certain use cases, it is desirable to have multiple instances of Cineast running on the same machine. At times the feature set is overlapping and as such, it is desirable to be able to configure these features in such a way that they can be shared across cineast instances (on the same machine) while simultaneously enabling the non-sharing of such features. In particular, service parameters should be freely definable.

PR #379, opened 2023-05-15
New External API Extraction Feature

silvanheller:
This Draft PR contains work towards #370 and additionally has some cleanup for feature-related code.

For testing purposes, there is a file named `external_extraction.json` which showcases the usage. This will be removed before merging.

## new features
 - new Feature `ExternalAPI` with configurable entity name, API endpoint, feature type, vector length, correspondence incl. parameters, distance, model.
   - currently expects endpoints to offer two things: `extract/image` which can handle base64-encoded pngs, and `extract/text`, which handles text. These enpoints can return a single vector or string which is then used for retrieval or persisted.

## cleanup
- unifying naming w.r.t entity / table. Now everything is named `entity` which is what we mostly use (esp. in Cottontail)


cineast-core/src/main/java/org/vitrivr/cineast/core/features/FESVectorFeature.java
```diff
@@ -414,6 +414,7 @@ private float[] embedModel(IModel model) throws IOException, InterruptedExceptio
 
   private float[] embedVideo(List<MultiImage> frames) throws IOException, InterruptedException {
     var images = frames.stream().map(image -> image.getBufferedImage()).collect(Collectors.toList());
+    images = images.stream().map(img -> new BufferedImage(img.getWidth(), img.getHeight(), BufferedImage.TYPE_INT_RGB)).collect(Collectors.toList());
```
lucaro, line 417: You are overwriting the previously generated images with empty images here, that is almost certainly not what you want.

PR #381, opened 2023-06-08
Optionally Have Segment Exporters Derive From AbstractSegmentExporter.

faberf:
For the upcoming features in the external-api-extraction branch, it would be useful to extract features from audio by sending a binary wav file data to the api. Exporting a wav file already exists in cineast, but the functionality was hard-coded to export to a local file and not an arbitrary data stream. To address this issue I refactored AudioSegmentExporter to inherit from AbstractSegmentExporter. This allows the programmer to only define how to export to a stream and then processSegment and other functions work automatically. I also implemented this on AudioSpectogramExporter as an example.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,12 +25,14 @@ COPY --from=build \
   /cineast-src/resources \
   /opt/cineast/resources
 
+WORKDIR /opt/cineast
+
 RUN printf '#!/bin/bash\n\
 if [ "$1" != "api" ] && [ "$1" != "cli" ]; then\n\
     echo "Usage: $0 api|cli" >&2\n\
     exit 1\n\
 fi\n\
-cd /opt/cineast/ && java -jar cineast-$1.jar ${@:2}'\
+java -jar "cineast-$1.jar" "${@:2}"'\
 > /opt/cineast/bootstrap.sh
 RUN chmod +x /opt/cineast/bootstrap.sh
 


cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,124 @@
...
+    try {
+      /* Prepare folder and OutputStream. */
+      Path folder = this.destination.resolve(shot.getSuperId());
+      if (!folder.toFile().exists()) {
+        folder.toFile().mkdirs();
+      }
+      Path path = folder.resolve(shot.getId() + this.fileExtension);
+      OutputStream os = Files.newOutputStream(path);
+
+      /* Write audio data to OutputStream. */
```
lucaro, line 74: This is data-type agnostic at this point, no?

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,124 @@
...
+abstract public class AbstractSegmentExporter implements Extractor {
+
+  protected static final Logger LOGGER = LogManager.getLogger();
+
+  private static final String PROPERTY_NAME_DESTINATION = "destination";
+
+  /**
+   * Destination path for the audio-segment.
+   */
+  private final Path destination;
```
sauterl, line 28: Why is this private object property documented and the following protected properties aren't?
I heavily suggest to document important information particularly on things that are exposed.
Our software stack has certainly room for improvement in terms of internal and external documentation and going forward, it seems to me it would be a good idea to aim for better documentation...

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Default constructor
+   */
+  public AbstractSegmentExporter() {
+    this(new HashMap<>());
+  }
+
+  /**
+   * Default constructor. The AudioSegmentExport can be configured via named properties in the provided HashMap.
```
sauterl, line 42: Copy-paster error: This isn't the `AudioSegmentExporter` anymore

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,124 @@
...
+   * <p>
+   * Supported parameters:
+   *
+   * <ol>
+   *      <li>destination: Path where files should be stored.</li>
```
sauterl, line 47: I would highlight in the documentation that its literally `destination`

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,124 @@
...
+   * </ol>
+   *
+   * @param properties HashMap containing named properties
+   */
+  public AbstractSegmentExporter(HashMap<String, String> properties) {
+    this.destination = Path.of(properties.getOrDefault(PROPERTY_NAME_DESTINATION, "."));
```
sauterl, line 53: I am not convinced that `.` is a sensible default here.
My suggestion would be `$(pwd)/export` to not fill the current working directory with individual files?

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  abstract public void exportToStream(SegmentContainer sc, OutputStream stream);
```
sauterl, line 56: Documentation missing, please add documentation

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,124 @@
...
+    exportToStream(shot, baos);
+
+    // convert the ByteArrayOutputStream's data to a byte array
+    byte[] bytes = baos.toByteArray();
+
+    // encode the byte array to a Base64 string
+    String base64 = Base64.getEncoder().encodeToString(bytes);
+
+    // concatenate the data URL prefix and the Base64 string
+    return this.dataUrlPrefix + base64;
```
sauterl, line 98: There is no guarantee at this point that `this.dataUrlPrefix` has been set

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,124 @@
...
+
+    // convert the ByteArrayOutputStream's data to a byte array
+    return baos.toByteArray();
+  }
+
+  @Override
+  public void init(PersistencyWriterSupplier phandlerSupply) { /* Nothing to init. */ }
+
+  @Override
+  public void finish() {  /* Nothing to finish. */}
```
sauterl, line 116: What about closing the stream?

cineast-core/src/main/java/org/vitrivr/cineast/core/features/exporter/AudioSpectogramExporter.java
```diff
@@ -78,30 +68,25 @@ public AudioSpectogramExporter() {
    * @param properties HashMap containing named properties
    */
   public AudioSpectogramExporter(HashMap<String, String> properties) {
-    this.destination = Paths.get(properties.getOrDefault(PROPERTY_NAME_DESTINATION, "."));
+    super(properties);
     this.width = Integer.parseInt(properties.getOrDefault(PROPERTY_NAME_WIDTH, "800"));
     this.height = Integer.parseInt(properties.getOrDefault(PROPERTY_NAME_HEIGHT, "600"));
     this.format = properties.getOrDefault(PROPERTY_NAME_FORMAT, "JPG");
+    this.fileExtension = "." + format.toLowerCase();
```
sauterl, line 75: So, format has to be a valid extension?

sauterl (changes requested):
Thanks for the PR. In my opinion there are some points to refactor / document. Also, since this change is particularly for the works in https://github.com/vitrivr/cineast/pull/379, I suggest to target that branch instead of main.

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
faberf: Can you please clarify this?

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
faberf: I agree with this and am fixing this. Just for the record, originally AudioSegmentExporter did use '.'

faberf:
Optionally refers to the fact that I am not making any breaking changes to existing exporters. They still work. If there is a need they can be refactored so that they derive from AbstractSegmentExporter and inherit additional functionality.

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,149 @@
...
+import org.vitrivr.cineast.core.db.PersistencyWriterSupplier;
+import org.vitrivr.cineast.core.db.setup.EntityCreator;
+import org.vitrivr.cineast.core.features.extractor.Extractor;
+import org.vitrivr.cineast.core.util.LogHelper;
+
+abstract public class AbstractSegmentExporter implements Extractor {
+
+  protected static final Logger LOGGER = LogManager.getLogger();
+
+  private static final String PROPERTY_NAME_DESTINATION = "destination";
```
sauterl, line 23: I'd prefer the property name to be public.

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
sauterl: I assume the path should be a folder? The documentation could be more precise.

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
sauterl: You have `destination` as the property key. I would expect that this is marked as such (e.g. with `@code` in line 57)

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
sauterl: To put emphasis on this comment. I'd expect the `finish` method be `abstract` since the implementing exporter should flush and close its stream, not?

cineast-core/src/main/java/org/vitrivr/cineast/core/features/abstracts/AbstractSegmentExporter.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  @Override
+  public void init(PersistencyWriterSupplier phandlerSupply) { /* Nothing to init. */ }
+
+  @Override
+  public void finish() {  /* Nothing to finish. */}
+
+  @Override
+  public void initalizePersistentLayer(Supplier<EntityCreator> supply) { /* Nothing to init. */ }
```
sauterl, line 144: Doesn't in your case mean persistent layer database + outgoing stream? In this case I'd expect this to be`abstract` such that the implementing exporter is able to performe proper initialisation.

sauterl (changes requested):
Please excuse  the late reply, this PR got lost in too many notifications. Thank you for the changes. I still think there is room for improvement.

Did you consider targeting the feature branch this work builds upon? In case you reject this suggestion I'd like to read about this in the discussion.

EDIT: Language

PR #382, opened 2023-06-09
[Docker] Add workdir

x4e-jonas:
Remove error prone `cd`


lucaro (approved):
Not sure why this would be error-prone, but it's certainly nicer to do it this way.

